PR #6, opened 2021-08-30
Add opentelemetry integration to Oban 

indrekj:
By default a new trace is automatically started when a job is processed
by monitoring these events:
* `[:oban, :job, :start]` — at the point a job is fetched from the database and will execute
* `[:oban, :job, :stop]` — after a job succeeds and the success is recorded in the database
* `[:oban, :job, :exception]` — after a job fails and the failure is recorded in the database

To also record a span when a job is created and to link traces together
`Oban.insert/2` has to be replaced by `OpentelemetryOban.insert/2`.

Before:

```elixir
  %{id: 1, in_the: "business", of_doing: "business"}
  |> MyApp.Business.new()
  |> Oban.insert()
```

After:

```elixir
  %{id: 1, in_the: "business", of_doing: "business"}
  |> MyApp.Business.new()
  |> OpentelemetryOban.insert()
```

### Checklist

**Acknowledgements**

- [x] I certify that I am authorized to assign ownership of this code to the OpenTelemetry project.
- [x] I acknowledge that ownership of the [Hex](https://hex.pm) package must be given/transferred to the [OpenTelemetry Hex Org](https://hex.pm/orgs/opentelemetry)

**Project Type**

- [x] Instrumentation
- [ ] Exporter
- [ ] Propagator
- [ ] Metapackage
- [ ] Example
- [ ] Util

**Prior Art**

- [ ] Is this an existing project? Includes private libraries being open sourced.
  - [ ] Yes
  - [x] No
- [ ] Are there any other packages on [Hex](https://hex.pm) providing similar functionality? - I'm not aware of it.

**Existing Project**

- [ ] Moving from an existing public repository? Provide a link in the description.
- [x] All authors who have contributed code contained in this PR have signed the [CNCF CLA](https://identity.linuxfoundation.org/projects/cncf) - no other authors at the moment

**All Projects**

- [x] Label created (maintainers)
- [x] [labeler.yml](https://github.com/open-telemetry/opentelemetry-erlang-contrib/blob/main/.github/labeler.yml) updated
- [x] CI
  - [x] Elixir or Erlang language file is updated
  - [x] Test strategy should use the matrix strategy with all supported language/OTP combinations
  - [x] The job is scoped to only run for relevant code changes
- [x] [Codeowners](https://github.com/open-telemetry/opentelemetry-erlang-contrib/blob/main/CODEOWNERS) adds scoped ownership for project maintainers
- [x] License file (Apache2 - copyright to OpenTelemetry Authors)
- [x] All code complies with the [Otel Specification](https://github.com/open-telemetry/opentelemetry-specification/tree/main/specification) and Otel Semantic Conventions - [Trace](https://github.com/open-telemetry/opentelemetry-specification/tree/main/specification/trace/semantic_conventions) [Metrics](https://github.com/open-telemetry/opentelemetry-specification/tree/main/specification/metrics/semantic_conventions)


instrumentation/opentelemetry_oban/test/opentelemetry_oban_test.exs
```diff
@@ -0,0 +1,168 @@
+defmodule OpentelemetryObanTest do
```
indrekj, line 1: There was a comment by @ricardoccpaiva before moving this to this repo:
> What do you think of using https://hexdocs.pm/oban/Oban.Testing.html for testing? I believe it's a better approach because it avoids using :telemetry.execute explicitly thus being more future proof in case of changes.

We thought that it's possible to use sqlite3 but it's not. Oban is PG only. I could try setting up PG but that's a bit more effort (especially in regards to CI).

bryannaegele:
Thanks for opening this @indrekj. Once I get the phoenix migration complete, I will try to write up a guide in the next few days on migrating a library to here.

I couldn't find the original repo out on github or the package in hex. Do you have links to those?

Also, one thing to get a jump on is verify anyone who has contributed to the original codebase has signed the CLA https://identity.linuxfoundation.org/projects/cncf.

indrekj:
> I couldn't find the original repo out on github or the package in hex. Do you have links to those?

I deleted the repository. Basically ported the initial (non-merged) PR to here. There have been no releases yet. That's why you cannot find the repo nor the package. I can release the first version (v0.1) after this gets merged.

> Also, one thing to get a jump on is verify anyone who has contributed to the original codebase has signed the CLA https://identity.linuxfoundation.org/projects/cncf.

Again, not an issue. There were no contributions before that. The current code is written by myself.

bryannaegele:
We'll do a formal review later as part of this, but I had a couple of general questions on the implementation that could be considered while we work out CI and processes.

* Propagation: There's currently no ability for propagating context to the jobs for linking spans to parents. In our implementation at work, we trace directly and aren't using telemetry. Is there a mechanism you can think of to propagate that ctx through job metadata that could be implemented?
* Job creation: We have a `process` event here but no corresponding `send` or `receive`. Are there mechanisms or opportunities here we could leverage?
* Batches: It would be nice to support batches with the batch id. This also relates to propagation in general.

indrekj:
> Propagation: There's currently no ability for propagating context to the jobs for linking spans to parents. In our implementation at work, we trace directly and aren't using telemetry. Is there a mechanism you can think of to propagate that ctx through job metadata that could be implemented?

I did look into this but didn't find a good transparent option. I initially thought that maybe using the Oban plugins mechanism could help here. But the jobs are inserted like [this](https://github.com/sorentwo/oban/blob/master/lib/oban.ex#L196-L221). This is basically a direct insertion to the database so there are no plugins involved.

There is a `meta` field on the job that could be used to propagate tracing context.

So if the user is willing to change the codebase then we could propagate the information through that.

E.g. before:
```
      %{account_id: 1, url: "https://example.com"}
      |> MyApp.Worker.new()
      |> Oban.insert()
```
after:
```
      %{account_id: 1, url: "https://example.com"}
      |> MyApp.Worker.new()
      |> OpentelemetryOban.enhance()
      |> Oban.insert()
```

where `OpentelemetryOban.enhance(worker)` would add propagation information to `worker.meta` field. Is this something you think would be a good solution?

> Job creation: We have a process event here but no corresponding send or receive. Are there mechanisms or opportunities here we could leverage?

`send`: The whole send is basically just an insert call to the Postgres database. So there aren't any hooks/plugins that we can use (at least I don't see them). Again, if we'd use `OpentelemetryOban.enhance()` helper method then theoretically we could create a span there. But that seems very brittle because it doesn't actually wrap the insert call and if the user inserts the job later then the timings will be wrong. It would be better to wrap or replace the insert call somehow, but I'm not exactly sure how. I'm open to ideas.

`receive`: Correct me if I'm wrong but to my understanding, this is meant to be used in case of batch processing or when receive and process and done separately. I checked a couple of ruby libraries ([1](https://github.com/open-telemetry/opentelemetry-ruby/blob/cc28b8648e4085ff62e65e3db87ff45f82750cd1/instrumentation/delayed_job/lib/opentelemetry/instrumentation/delayed_job/plugins/tracer_plugin.rb#L34), [2](https://github.com/open-telemetry/opentelemetry-ruby/blob/cc28b8648e4085ff62e65e3db87ff45f82750cd1/instrumentation/resque/lib/opentelemetry/instrumentation/resque/patches/resque_job.rb#L32), [3](https://github.com/open-telemetry/opentelemetry-ruby/blob/cc28b8648e4085ff62e65e3db87ff45f82750cd1/instrumentation/sidekiq/lib/opentelemetry/instrumentation/sidekiq/middlewares/server/tracer_middleware.rb#L26)) and they all seemed to use `process` without the `receive` event. Oban seems to support [batch workers](https://hexdocs.pm/oban/batch.html) but that's available only through the Pro version. I'd prefer to focus on the free version first.

bryannaegele:
> So if the user is willing to change the codebase then we could propagate the information through that.

I would strongly suggest supporting that and selection of metadata keys for inclusion as attributes.

This is from `0.4` code, but we have this in our workers, but with the new span ctx format, I'd recommend using `text_map_inject`. Oban serializes metadata to json, so it's easier to just use propagators.

```elixir
  defp decode_ctx("undefined"), do: :undefined

  defp decode_ctx(ctx) do
    Enum.map(ctx, fn
      el when is_binary(el) -> String.to_existing_atom(el)
      el -> el
    end)
    |> List.to_tuple()
  end
```

indrekj:
@bryannaegele I did a bunch of changes here.

But basically now if the user uses `OpentelemeryOban.insert/2` instead of `Oban.insert/2` then a `send` span is created and also the context is propagated through the Oban job metadata. On the receiving side, we create a new trace and link them together.

Also added some instructions to the project README file.

I'll deal with CI after opentelemetry_phoenix gets merged to avoid duplicate work. The tests pass locally, but for CI (Github actions) I need to somehow set up Postgres.

tsloughter:
> But basically now if the user uses `OpentelemeryOban.insert/2` instead of `Oban.insert/2` then a `send` span is created and also the context is propagated through the Oban job metadata. On the receiving side, we create a new trace and link them together.

Nice!

And sadly this is the reality of libraries using `telemetry` for instrumentation related to tracing :(, especially if they don't offer any middleware/hooking functionality.

`telemetry` is filling the role that `opentelemetry_api` is meant to, but with less functionality -- the "role" being a way for everyone to instrument their code with a common library that by default does nothing, so users who don't want it can ignore it, and is completely pluggable in what it does do if the user chooses to enable it.

Not that I don't think that was completely reasonable to do :), just stating the state of things.

There is likely a need for a push for libraries that use `telemetry` and make these sort of external service calls to all include a pluggable way to extend those calls. Then when `opentelemetry_oban` starts it can register a function based on the configured opentelemetry propagators.

Because some users who want to enable tracing are going to be using Oban indirectly and won't be able to simply use the new function. Same issue exists for HTTP, database, MQ, Kafka, etc clients

dvic:
I just came across this PR, FYI we have also an implementation available at https://github.com/qdentity/opentelemetry_oban. For this I had to get two PRs into Oban: https://github.com/sorentwo/oban/pulls?q=is%3Apr+author%3Advic+is%3Aclosed. We can have a look and, if needed, merge missing pieces into this PR.

indrekj:
@tsloughter Rebased my PR and addressed the checklist in the open checklist PR.

@dvic I did see that repository, though only after most of my code was already done. I would like to get this PR approved/merged before porting over your changes. This current PR takes care of job processing and trace propagation. I see you added tracing for circuits, plugins etc. It likely makes sense to port them over but as it needs some testing and making sure they follow semantic conventions I'd like to start with the basic features that this PR should take care of.

dvic:
@indrekj I just tried your PR and I must say, it's looking good :) I like the way you organized the attributes and distinguished between process/send spans, while also supporting linking them to each other. I archived https://github.com/qdentity/opentelemetry_oban and referred to this PR.

Let me know if there's anything I can do to get this PR merged and published 👍

dvic:
Another thing that you could add this PR: `OpentelemetryOban.insert!/1`, I noticed that one was missing (Oban has this one)

indrekj:
Ack, I'll take a look tomorrow

dvic:
> Ack, I'll take a look tomorrow

👍 Another thing I just encountered that you might want tackle as well: https://github.com/ricardoccpaiva/opentelemetry_tesla/pull/5 (it seems we have to do `&__MODULE__.my_handle_fun/4` instead of ``&my_handle_fun/4`)

indrekj:
@dvic Updated

dvic:
> @dvic Updated

Perfect, thanks 👍

instrumentation/opentelemetry_oban/lib/opentelemetry_oban.ex
```diff
@@ -0,0 +1,121 @@
...
+  def insert_all(name \\ __MODULE__, multi, multi_name, changesets_or_wrapper) do
+    Oban.insert_all(name, multi, multi_name, changesets_or_wrapper)
+  end
+
+  defp add_tracing_information_to_meta(changeset) do
+    meta = Changeset.get_field(changeset, :meta, %{})
+
+    new_meta =
+      []
+      |> :otel_propagator.text_map_inject()
```
indrekj, line 96: New OTEL version in the main branch changes this to: `:otel_propagator_text_map.inject`

dvic:
> All the child spans also get the same sampling decision. So the ecto queries are not recorded either.
> 
> I think this is a more convenient option than suggesting users to create a separate repo for Oban.

Ah yes of course, this works because of the explicit "sampling off". By the way, before you start using this in prod: I fixed a memory leak (see https://github.com/open-telemetry/opentelemetry-erlang/pull/267) that I stumbled upon when doing this myself.

> In case you were using the previous version (the one without pluings part introduced) then that's still on hex (0.1.0-rc). I marked this PR as WIP as I'm using things from otel API rc3 which is not released yet but should be any day now.

👍

derekkraan:
@bryannaegele I looked at this PR, it appears to be following the conventions laid out in https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/messaging.md.

Span names all check out (eg: `"#{worker} send"`, `"#{worker} process"`). Attribute names also seem to be conform the spec, eg:
```
      "messaging.system": "oban",
      "messaging.destination": queue,
      "messaging.destination_kind": "queue",
```

dvic:
@indrekj are you aware that the `sampler` option has been removed from `start_span`? I was already wondering why all the internal oban spans started showing up :) 

@tsloughter any updates on this issue? Are there any other options currently for a library to ignore particular child spans (in this case Ecto queries)?

tsloughter:
@dvic I'm missing, or forgetting, some context, there are Ecto queries you don't want included as child spans of an Oban span?

instrumentation/opentelemetry_oban/lib/opentelemetry_oban/job_handler.ex
```diff
@@ -0,0 +1,102 @@
...
+        if(inserted_at, do: DateTime.to_iso8601(inserted_at), else: nil),
+      "messaging.oban.scheduled_at": DateTime.to_iso8601(scheduled_at)
+    ]
+
+    span_name = "#{worker} process"
+
+    OpentelemetryTelemetry.start_telemetry_span(@tracer_id, span_name, metadata, %{
+      kind: :consumer,
+      links: links,
+      sampler: sampler
```
bryannaegele, line 79: Sampling at the span level is no longer supported so you can safely remove any references to sampling. Users will have to run their own sampler to cover these spans.

I'm not sure if there is a pattern we could establish for instrumentation libraries to provide their own samplers a user could employ. Maybe @tsloughter has an idea on that as it seems like a common desire.

instrumentation/opentelemetry_oban/lib/opentelemetry_oban/job_handler.ex
```diff
@@ -0,0 +1,102 @@
...
+        scheduled_at: scheduled_at,
+        attempt: attempt,
+        max_attempts: max_attempts,
+        meta: job_meta
+      }
+    } = metadata
+
+    :otel_propagator_text_map.extract(Map.to_list(job_meta))
+    parent = OpenTelemetry.Tracer.current_span_ctx()
+    links = if parent == :undefined, do: [], else: [OpenTelemetry.link(parent)]
```
bryannaegele, line 56: It may be worth noting in docs that linking is not supported by a lot of tooling, especially commercial tools.

.github/workflows/elixir.yml
```diff
@@ -104,3 +104,39 @@ jobs:
...
+    env:
+      app: 'opentelemetry_oban'
+    defaults:
+      run:
+        working-directory: instrumentation/${{ env.app }}
+    runs-on: ubuntu-18.04
+    name: Opentelemetry Oban test on Elixir ${{ matrix.elixir_version }} (OTP ${{ matrix.otp_version }})
+    strategy:
+      fail-fast: false
+      matrix: ${{ fromJson(needs.test-matrix.outputs.matrix) }}
```
bryannaegele, line 119: I don't see any setup of postgres for your tests. You can copy from the ecto tests.

https://github.com/open-telemetry/opentelemetry-erlang-contrib/pull/6/files#diff-567a3fdf49d6f0f0119dda6d1c098fcb5ee3785ee956481b3daaf5a2f246076fR37-R45

instrumentation/opentelemetry_oban/lib/opentelemetry_oban/job_handler.ex
dvic: I believe Honeycomb does support this, I see nice links between the process and send spans :)

dvic:
> @dvic I'm missing, or forgetting, some context, there are Ecto queries you don't want included as child spans of an Oban span?

I've explained it poorly :) This PR creates spans for Oban jobs but also for internal events of Oban (currently only one particular kind of "plugin" events). The Oban plugin events result in various child Ecto spans if you use opentelemetry_ecto. These Ecto spans were not sampled by default because the sampler option was set to "always off" for the parent Oban plugin spans (created by this library). Now that this option is not supported anymore, I wonder if there is a recommended way for libraries to provide "building blocks/helpers" for disabling these particular spans.

tsloughter:
@dvic ah, ok. But no, there is no solution to this at the moment.

instrumentation/opentelemetry_oban/lib/opentelemetry_oban/job_handler.ex
tsloughter: There isn't. It should be possible to configure per instrumentation library (tracer) but isn't yet, see my OTEP https://github.com/open-telemetry/oteps/pull/186

tsloughter:
> I discovered something fun in Oban. The telemetry event `[:oban, :job, :exception]` comes some of the time from the job process, but if a nested process has died (so you did something like `Task.start_link` in your Oban job), then it comes from _another_ process (https://github.com/sorentwo/oban/blob/v2.9.2/lib/oban/queue/producer.ex#L90). This causes the telemetry handler to die and detach(!!) because the span is not present in the process dictionary of the monitoring process. We solved this by using a monitor (see: [open-telemetry/opentelemetry-erlang#84](https://github.com/open-telemetry/opentelemetry-erlang/pull/84)), not listening to `[:oban, :job, :exception]` at all

Eek...

We should probably go ahead and merge this once the `sampler` stuff is removed, but that sounds icky. I don't understand why is the event being run in a different process? If you were saying if the worker process dies it results in the `exception` event happening in a process without the context I'd understand, but if a process started by the worker process dies why is it different?

indrekj:
@tsloughter @bryannaegele Updated

* Sampler has been removed
* CI is now green - Postgres configuration was indeed missing
* Span linking - I would like to add a `propagation_style` configuration option so the user can choose between Links and Child-Of style. Similar to Ruby [libraries](https://github.com/open-telemetry/opentelemetry-ruby/blob/9c031a1bece68c1f40fca17a1cc17cd1e784818e/instrumentation/que/lib/opentelemetry/instrumentation/que/instrumentation.rb#L30-L47). But I would like to do that in a separate PR after this PR, because this PR is already huge and I don't want to make it even bigger.

dvic:
> @tsloughter @bryannaegele Updated
> 
> * Sampler has been removed
> * CI is now green - Postgres configuration was indeed missing
> * Span linking - I would like to add a `propagation_style` configuration option so the user can choose between Links and Child-Of style. Similar to Ruby [libraries](https://github.com/open-telemetry/opentelemetry-ruby/blob/9c031a1bece68c1f40fca17a1cc17cd1e784818e/instrumentation/que/lib/opentelemetry/instrumentation/que/instrumentation.rb#L30-L47). But I would like to do that in a separate PR after this PR, because this PR is already huge and I don't want to make it even bigger.

Great :) Should we  add to the docs a line about how to exclude the internal query (by tracing the plugin events and then using a root sampler)?

tsloughter (approved):

tsloughter:
Agreed on a separate PR for adding propagation style.

tsloughter (approved):

dvic (approved):

dvic:
I have used the following sampler:

```elixir
defmodule MySampler do
  @behaviour :otel_sampler

  defstruct [:opts]

  @impl true
  def setup(opts), do: %__MODULE__{opts: opts}

  @impl true
  def description(%__MODULE__{}), do: "MySampler"

  @impl true
  def should_sample(ctx, _trace_id, _links, span_name, _span_kind, attributes, _opts) do
    drop =
      String.ends_with?(span_name, ".repo.query:oban_producers") ||
        String.ends_with?(span_name, ".repo.query:oban_jobs") ||
        String.starts_with?(span_name, "Elixir.Oban.Plugins") ||
        String.starts_with?(span_name, "Elixir.Oban.Pro.Plugins") ||
        Keyword.get(attributes, :"db.statement") in ["begin", "commit"]

    decision =
      if drop do
        :drop
      else
        :record_and_sample
      end

    attributes = []

    tracestate =
      OpenTelemetry.Tracer.current_span_ctx(ctx)
      |> OpenTelemetry.Span.tracestate()

    {decision, attributes, tracestate}
  end
end
```

and then in the config:

```elixir
my_sampler = {MySampler, []}

config :opentelemetry,
  sampler:
    {:parent_based,
     %{
       root: my_sampler,
       local_parent_sampled: my_sampler,
       remote_parent_sampled: my_sampler
     }}
```

If you think it's useful we could add this to the docs as a starter.

indrekj:
I'm personally not a fan of that approach. It requires the user to manually list every possible span there could be. If the underlying library changes and there's a new span then this has to be modified again. This can be quite cumbersome when there are many services.

In general, this seems to be a common problem and I think API/SDK should provide a way to start a non-recording span. E.g. In ruby there's a helper method `[untraced](https://github.com/open-telemetry/opentelemetry-ruby/blob/2bf0e6aed060867bc65560a92e2c464d48b26a02/common/lib/opentelemetry/common/utilities.rb#L75-L77)` for instrumentation libraries. These are used in libraries like [sidekiq](https://github.com/open-telemetry/opentelemetry-ruby/blob/main/instrumentation/sidekiq/lib/opentelemetry/instrumentation/sidekiq/patches/poller.rb#L19).

dvic:
> I'm personally not a fan of that approach. It requires the user to manually list every possible span there could be. If the underlying library changes and there's a new span then this has to be modified again. This can be quite cumbersome when there are many services.
> 
> In general, this seems to be a common problem and I think API/SDK should provide a way to start a non-recording span. E.g. In ruby there's a helper method `[untraced](https://github.com/open-telemetry/opentelemetry-ruby/blob/2bf0e6aed060867bc65560a92e2c464d48b26a02/common/lib/opentelemetry/common/utilities.rb#L75-L77)` for instrumentation libraries. These are used in libraries like [sidekiq](https://github.com/open-telemetry/opentelemetry-ruby/blob/main/instrumentation/sidekiq/lib/opentelemetry/instrumentation/sidekiq/patches/poller.rb#L19).

Interesting, but children of a non-recording span are still recording right? How would this pan out with the ecto queries problem?

indrekj:
> Interesting, but children of a non-recording span are still sampled right?

No, they're not. The decision is propagated to all the children spans.

dvic:
> > Interesting, but children of a non-recording span are still sampled right?
> 
> No, they're not. The decision is propagated to all the children spans.

Ah I see, didn't know that, thanks!

tsloughter:
@indrekj so you've got the tests against postgres working in CI? Making this good to merge now?

indrekj:
Yup I think so

instrumentation/opentelemetry_oban/lib/opentelemetry_oban/job_handler.ex
```diff
@@ -0,0 +1,101 @@
...
+      "messaging.oban.scheduled_at": DateTime.to_iso8601(scheduled_at)
+    ]
+
+    span_name = "#{worker} process"
+
+    OpentelemetryTelemetry.start_telemetry_span(@tracer_id, span_name, metadata, %{
+      kind: :consumer,
+      links: links
+    })
+    |> Span.set_attributes(attributes)
```
dvic, line 80: @indrekj why is this done with `Span.set_attributes`? It's not possible to set the attributes directly in the `start_opts` here? The reason why I'm asking is that I don't see the attributes in my sampler (but they do appear in Honeycomb). I'm receiving this span without the attributes in my sampler. @tsloughter is this intended behaviour of the sampler / span design?

indrekj: I think I used some other project as an example. The `start_opts` works with `attributes` key indeed: https://github.com/open-telemetry/opentelemetry-erlang-contrib/compare/9ff9f64044d47306c0764dd9e12472a1b7eb3349..5b50df46ca8c0847f89e595cf74d4a1e2892c6e8 . I'm also using set_attributes in two other places (insert functions) but there I don't have access to all attributes immediately so there it makes sense to use set_attributes.

dvic: With this change now I do get the attributes in the sampler, but I'm not sure if this is a bug or not, I was expecting this to work also with `Span.set_attributes`.

indrekj: I think it **might** be intentional. Imagine you have this flow:

1. Start span A
2. Send outbound request to service B
3. Set attributes to span A
4. Do something and finish span A

In the step 2, the sampling decision has to be known. Otherwise, service B doesn't know if the spans should be sampled or not. So the sampler has to be run in the step 1.

instrumentation/opentelemetry_oban/lib/opentelemetry_oban/job_handler.ex
```diff
@@ -0,0 +1,101 @@
...
+        attempt: attempt,
+        max_attempts: max_attempts,
+        meta: job_meta
+      }
+    } = metadata
+
+    :otel_propagator_text_map.extract(Map.to_list(job_meta))
+    parent = OpenTelemetry.Tracer.current_span_ctx()
+    links = if parent == :undefined, do: [], else: [OpenTelemetry.link(parent)]
+    OpenTelemetry.Tracer.set_current_span(:undefined)
```
dvic, line 57: @indrekj Due to the force pushes on this PR I lost a bit track when this was introduced (maybe it's a good idea to do one force push at the end ;)), but currently I'm having the problem that the span started inside a job are not children of the job span, is this intended? Are you also seeing this behaviour? The consequence of this is that my custom sampler is not that useful anymore :)

dvic: I mean, I can always start another wrapper span inside the job, but I just wanted to double check and understand that this is intended behavior.

indrekj: We're not seeing this. I also added a test to ensure it: https://github.com/open-telemetry/opentelemetry-erlang-contrib/compare/db0e1f1a67bbd51992b10a0020da3b21119f6edb..ddf26187cee991e329c9dcbeaa42ad50eacc0c5d

We however had a similar issue when we were using `def timeout(_job) do` in the jobs. If timeout function is defined then oban uses perform_timed to execute the worker, see: https://github.com/sorentwo/oban/blob/60274b76ff2c59d660ae1bf7721a3ff14854700b/lib/oban/queue/executor.ex#L257-L269 . This creates a new process so the current span information is lost. We only had one job in the entire system with that method, so we used:

```
  # We use our own `with_timeout` instead of Oban.Worker.timeout/1 because if
  # the latter is used then the trace context is not propagated.
  defp with_timeout(fun) do
    task = Task.async(OpentelemetryFunction.wrap(fun))

    case Task.yield(task, @timeout) || Task.shutdown(task) do
      {:ok, reply} ->
        reply

      nil ->
        {:error, :timeout}
    end
  end
```

indrekj: I'd like to get this PR merged to start addressing these kind of rare edge cases. As you said, the PR is huge and it's hard to follow the changes now.

dvic: Ahhh, yes in this case we also used `Oban.Worker.timeout/1`, thanks! 👍 

> I'd like to get this PR merged to start addressing these kind of rare edge cases. As you said, the PR is huge and it's hard to follow the changes now.

I agree, maybe we can track these open issues / edge cases in the description so we can work on it when the PR is merged.

tsloughter: Yea, I can merge and then you should start opening issues.

I saw some question about `set_attributes` and sampling but can't find it again, so answering here: The sampler won't see attributes set with `set_attributes` because it has already been sampled at that point. A span is started and the attributes passed to the function that starts the span will be available but any attributes set after starting the span are of course not available.

Not sure what the code was, maybe it was something that used a pipe to look like it was using the "builder pattern" to construct a span? But we don't use that pattern, so there is no "build" function that would come after doing something like `span("name") |> set_attributes(...) |> start()`. So yea, must pass any attributes that need to be available for sampling to the start call.

indrekj: Thanks. It was here: https://github.com/open-telemetry/opentelemetry-erlang-contrib/pull/6/files/9ff9f64044d47306c0764dd9e12472a1b7eb3349#diff-ca8cd6f4f23c284564be4b34cb5caf868177e4190300720bc3c6fdb255c18b90 . For that particular case we were able to avoid set_attributes

> Yea, I can merge and then you should start opening issues.

:+1:

indrekj:
PR has been approved for some time now, and there doesn't seem to be more comments. Could somebody merge it?

tsloughter:
Thanks for the persistence :). I keep preparing to merge it, hit the update button, checks start running again... and I move on to something else and forget :(. Checks are again running after I hit the `update` button, this time I'll try not to move on to something else.

tsloughter:
Merged.
 
I looked up the package on hex and it appears to be owned by a company/org "salemove". Is that you?

We need the opentelemetry org/team added to it on hex so we can also publish if need be -- as in, we won't be publishing the new release but in case the other owner went away or wasn't available and there was a security vulnerability, stuff like that, we'd need to be able to publish.

indrekj:
@tsloughter Thx.

Ownership has been transferred to `opentelemetry` now. See https://hex.pm/packages/opentelemetry_oban . I think I lost all permissions myself now, but that doesn't really matter I think. Probably releasing new versions should be part of the github actions flow anyway.

tsloughter:
Thanks, you should stay as an owner to be able to publish. It is much better to release manually and not github actions. Unless github actions has a way to require manual intervention during an action which may make it a viable option.

indrekj:
> Thanks, you should stay as an owner to be able to publish.

Could you run `mix hex.owner add opentelemetry_oban indrek` in that case? All my permissions were lost when I transferred this between organizations.

tsloughter:
Hm, that didn't work, but I swear you can have both a user and an owner as the owner. It returned "cannot add owner that is not a member of the organization"

PR #9, opened 2021-09-12
Update otel-phoenix deps

bryannaegele:
Update deps for opentelemetry-phoenix, test matrix versions, and address some feedback from the merge PR.


instrumentation/opentelemetry_phoenix/mix.exs
```diff
@@ -56,9 +58,9 @@ defmodule OpentelemetryPhoenix.MixProject do
   defp deps do
     [
       {:opentelemetry_api, "~> 1.0.0-rc"},
-      {:opentelemetry, "~> 1.0.0-rc"},
       {:opentelemetry_telemetry, "~> 1.0.0-beta"},
-      {:telemetry, "~> 0.4"},
+      {:telemetry, "~> 0.4 or ~> 1.0.0"},
```
indrekj, line 62:
```suggestion
      {:telemetry, "~> 0.4 or ~> 1.0"},
```
I don't think we need to lock the minor version

bryannaegele: It's safer to pin to minor as we may later introduce something that is reliant on something added in a version > 1.1. Your proposal would allow 0.4 - 2.0.

indrekj: My proposal would allow 0.4 - 1.[x] (2.0 excluded). You could always define the minimum minor version like `~> 1.1`. This would still allow users to use `1.2`, `1.3`, etc. Generally, I think the unlocked minor version is better, because if's locked and one library requires `1.0` and another one requires `1.1` then the user cannot use them together. This is also how it is done in the phoenix library as well: https://github.com/phoenixframework/phoenix/blob/473b10bb4166c57700b1e6d1a518f76458640da7/mix.exs#L66 .

Just my thoughts.

PR #14, opened 2021-09-17
Add support for sampling in opentelemetry_phoenix

dvic:
This PR has been moved from https://github.com/opentelemetry-beam/opentelemetry_phoenix/pull/32


dvic:
> Yea, it is only at the config level. Sadly this is also only a single sampler for all traces too. I'm working on an otep to allow a sampler to be set for a specific library that will resolve this problem (you'll be able to configure the tracer for `opentelemetry_phoenix` for example).

Alright, then this can be closed for now. Thanks!

PR #24, opened 2021-10-13
Upgrade Phoenix integration to otel rc.3

dvic:


dvic:
Submitted a PR in https://github.com/open-telemetry/opentelemetry-erlang/pull/290

indrekj:
> we are like minded I guess haha

I guess indeed. I should have double checked first.

> I'm getting warnings locally

I'm using b3 propagation, that's why I didn't notice that. I think your fix makes sense though.

PR #27, opened 2021-10-13
Otel 1.0.0-rc.3 support

bryannaegele:
1.0.0-rc3 support.

Also discovered with something in the recent changes that cowboy was hanging onto a current context between tests when a propagated context was extracted. I have to assume this is a cowboy reuse of processes or test setup oddity since the context shouldn't be set on a newly created process. The simple fix was to just clear the ctx at the end of requests.


dvic:
@bryannaegele Just want to call to attention the following: https://github.com/open-telemetry/opentelemetry-erlang/pull/290, I think it's wise to have this merged as rc.4 and use that, otherwise you get a log entry for every Phoenix request?

tsloughter (approved):

PR #40, opened 2021-11-18
inspect terms that otherwise can't be serialized by exporter

tverlaan:
When you have a case clause error or similar and the argument was a map or tuple for example it won't show up in honeycomb. This data is extremely useful to understand the actual reason of the case clause error.

Do let me know if there is a better place to fix this.


tsloughter (approved):
Looks right. Though the CI seems broken.

bryannaegele (approved):

PR #41, opened 2021-11-19
Set error status for ecto errors in opentelemetry_ecto

hoyon:
Right now when a query has an error (eg invalid field or timeout) the error attribute is set but not the opentelemetry span status or span status message.

This causes errors to not be marked as errors in datadog.

I've also tested this in jaeger which supported the previous attribute and also the replacement span statuses.


instrumentation/opentelemetry_ecto/lib/opentelemetry_ecto.ex
```diff
@@ -100,6 +93,14 @@ defmodule OpentelemetryEcto do
         attributes: attributes ++ base_attributes
       })
 
+    case query_result do
+      {:error, error} ->
+        OpenTelemetry.Span.set_status(s, OpenTelemetry.status(:error, Postgrex.Error.message(error)))
```
andrewhr, line 98: [The `Postgrex.Error` module is an exception struct](https://github.com/elixir-ecto/postgrex/blob/0f0e70080e55dceca4c0cc6e7a1b81ca835c49b2/lib/postgrex/error.ex#L2), so changing to [Exception.message/1](https://hexdocs.pm/elixir/1.12/Exception.html#message/1) makes this code agnostic to database adapters.

I'm not 100% all errors captured here are Exceptions, though.

hoyon: I've updated the code to use `Exception.message/1` if it's an exception.

bryannaegele (approved):

PR #44, opened 2021-11-23
Remove exporter race-condition on tests

andrewhr:
Default exporter immediately attempts on start connect to `:otel-collector` default port. As we don't have any collector running on our test environment, this results in a few warnings. That's not an issue in itself, as code imperially switches to another export, but creates a lot of noise.

This patch moves the exporter setup to `config/test.exs`, essentially removing the need to restart `:opentelemetry` application for each test case. The only work setup blocks do is update the exporter's target pid.

The processor was changed to simple mode, available now, which also remove another vector of (unlikely but theoretically possible) race-conditions.


tsloughter:
Looks good, sorry so late getting tot his, can you rebase your changes on the latest?

bryannaegele (approved):

Issue #36, opened 2021-11-16
cowboy integration always sets error on span with http.status set to 0


dvic:
I debugged this and it turns out `resp_status` is not an integer but `"200 OK"` in my case (`cowboy` 2.9.0 and `cowboy_telemetry` 0.3.1), while the code (as far as I can see) assumes it's an integer?

According to https://github.com/ninenines/cowboy/blob/8795233c57f1f472781a22ffbf186ce38cc5b049/src/cowboy.erl#L37 the resp_status can indeed be a binary as well.

Furthermore, shouldnt' `http.status` be `http.status_code`?

bryannaegele:
Resolved by #48

PR #48, opened 2021-11-30
Handle binary and undefined response status and rename http.status to http.status_code in `opentelemetry-cowboy`

### LINKED ISSUE #35: Error on adding `opentelemetry_cowboy` to Phoenix app
**Describe the bug**
Adding `:opentelemetry_cowboy.setup()` to the `start/2` function in `application.ex` in a Phoenix app, results in an error message on server start.

**Expected behavior**
No error and Cowboy spans emitted.

**Additional context**
Versions used:

```
{:opentelemetry, "1.0.0-rc.3"}
{:opentelemetry_api, "1.0.0-rc.3.2"}
{:opentelemetry_cowboy, "0.1.0"}
```

Also reproduced by @maciej-szlosarczyk on an empty Phoenix project: https://gist.github.com/maciej-szlosarczyk/91c257b738c429784889fb579f6e3636

kw7oe:
Attempt to resolve #35 and #36.

This PR includes the following changes:

- Rename `http.status` to `http.status_code` according to the [OpenTelemetry Specification](https://github.com/open-telemetry/opentelemetry-specification/blob/01169ed5fec0107afffb7bb5d8ea5466ae7241a0/specification/trace/semantic_conventions/http.md#common-attributes).
- Handle Cowboy might return binary response status as reported in #36.
  - **Reference:** Cowboy [`http_status` type spec](https://github.com/ninenines/cowboy/blob/2a082504996afccf78185182cd168123800bd4f0/src/cowboy.erl#L37)
- Handle Websocket upgrade request that will have `undefined` status but no error information as reported in #35.


instrumentation/opentelemetry_cowboy/src/opentelemetry_cowboy.erl
```diff
@@ -57,16 +57,22 @@ handle_event([cowboy, request, stop], Measurements, Meta, _Config) ->
...
-        Status when Status >= 400 ->
-            otel_span:set_attributes(Ctx, [{'http.status', Status}]),
+          case maps:get(error, Meta, undefined) of
+            {ErrorType, Error, Reason} ->
+              otel_span:add_event(Ctx, atom_to_binary(ErrorType, utf8), [{error, Error}, {reason, Reason}]),
+              otel_span:set_status(Ctx, opentelemetry:status(?OTEL_STATUS_ERROR, Reason));
+            _ ->
+              % do nothing first as I'm unsure how should we handle this
+              ok
+          end;
```
kw7oe, line 70: I'm not sure how would we want to handle the other scenario where we might have `undefined` status code but is not an error. As far as I know, it could be a websocket connection based on the stacktrace below:

```
Reason={:badkey, :error}
Stacktrace=[
  {:erlang, :map_get,
   [
     :error,
     %{
       pid: #PID<0.708.0>,
       ref: TesteWeb.Endpoint.HTTP,
       req: %{
         body_length: 0,
         cert: :undefined,
         has_body: false,
         headers: %{
           "accept" => "*/*",
           "accept-encoding" => "gzip, deflate",
           "accept-language" => "en-US,en;q=0.5",
           "cache-control" => "no-cache",
           "connection" => "keep-alive, Upgrade",
           "cookie" => "user_data=eyJoZXhfY29sb3IiOiIjNjBBNUZBIiwibmFtZSI6bnVsbH0=; _teste_key=SFMyNTY.g3QAAAABbQAAAAtfY3NyZl90b2tlbm0AAAAYTnM1Q254VGhPMWRva2hBem1zOFJKcmVO.-cfAkiIzqtt3BPMkQYFfNZWOd-iijLzaNEz1CAsly0o",
           "host" => "localhost:4000",
           "origin" => "http://localhost:4000",
           "pragma" => "no-cache",
           "sec-fetch-dest" => "websocket",
           "sec-fetch-mode" => "websocket",
           "sec-fetch-site" => "same-origin",
           "sec-websocket-extensions" => "permessage-deflate",
           "sec-websocket-key" => "MUMOmAyxUWbViEGZl/pJvA==",
           "sec-websocket-version" => "13",
           "upgrade" => "websocket",
           "user-agent" => "Mozilla/5.0 (Macintosh; Intel Mac OS X 10.15; rv:94.0) Gecko/20100101 Firefox/94.0"
         },
         host: "localhost",
         method: "GET",
         path: "/phoenix/live_reload/socket/websocket",
         peer: {{127, 0, 0, 1}, 53051},
         pid: #PID<0.708.0>,
         port: 4000,
         qs: "vsn=2.0.0",
         ref: TesteWeb.Endpoint.HTTP,
         scheme: "http",
         sock: {{127, 0, 0, 1}, 4000},
         streamid: 1,
         version: :"HTTP/1.1"
       },
       resp_headers: :undefined,
       resp_status: :undefined,
       streamid: 1
     }
   ], [error_info: %{module: :erl_erts_errors}]},
```

bryannaegele: I'm not sure what to do, either. I would have thought a websocket upgrade request would return `101` on `resp_status` if this stacktrace was triggered in a stop event. I'm not well-versed in how cowboy manages that sequence.

instrumentation/opentelemetry_cowboy/src/opentelemetry_cowboy.erl
```diff
@@ -57,16 +57,22 @@ handle_event([cowboy, request, stop], Measurements, Meta, _Config) ->
...
+    case StatusCode of
         undefined ->
-            {ErrorType, Error, Reason} = maps:get(error, Meta),
-            otel_span:add_event(Ctx, atom_to_binary(ErrorType, utf8), [{error, Error}, {reason, Reason}]),
-            otel_span:set_status(Ctx, opentelemetry:status(?OTEL_STATUS_ERROR, Reason));
-        Status when Status >= 400 ->
-            otel_span:set_attributes(Ctx, [{'http.status', Status}]),
+          case maps:get(error, Meta, undefined) of
+            {ErrorType, Error, Reason} ->
+              otel_span:add_event(Ctx, atom_to_binary(ErrorType, utf8), [{error, Error}, {reason, Reason}]),
```
bryannaegele, line 65: `atom_to_binary` should not be needed. That was an otel bug

bryannaegele:
@kw7oe does this resolve #61?

kw7oe:
> @kw7oe does this resolve #61?

Yeap, but the PR probably only cover an `Upgrade` connection, while this solve the other unforeseen cases where there's no error when the status is undefined.

instrumentation/opentelemetry_cowboy/src/opentelemetry_cowboy.erl
kw7oe: updated it after merge master 👍🏻

instrumentation/opentelemetry_cowboy/src/opentelemetry_cowboy.erl
```diff
@@ -57,16 +57,22 @@ handle_event([cowboy, request, stop], Measurements, Meta, _Config) ->
...
+              otel_span:set_status(Ctx, opentelemetry:status(?OTEL_STATUS_ERROR, Reason));
+            _ ->
+              % do nothing first as I'm unsure how should we handle this
+              ok
+          end;
+        StatusCode when StatusCode >= 400 ->
+            otel_span:set_attributes(Ctx, [{'http.status_code', StatusCode}]),
             otel_span:set_status(Ctx, opentelemetry:status(?OTEL_STATUS_ERROR, <<"">>));
```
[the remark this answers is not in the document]
kw7oe: @tsloughter you are referring to this part right:

> For HTTP status codes in the 4xx range span status MUST be left unset in case of SpanKind.SERVER and MUST be set to Error in case of SpanKind.CLIENT.

I have removed it from that clause. However, I also notice the following line:

> For HTTP status codes in the 5xx range, as well as any other code the client failed to interpret, span status MUST be set to Error.

This mean that we would need to set the status as error for `StatusCode >= 500` right?

instrumentation/opentelemetry_cowboy/test/opentelemetry_cowboy_SUITE.erl
```diff
@@ -9,8 +9,6 @@
 -include_lib("opentelemetry/include/otel_span.hrl").
 -include_lib("opentelemetry_api/include/otel_tracer.hrl").
 
--define(assertListsMatch(List1, List2), ?assertEqual(lists:sort(List1), lists:sort(List2))).
```
kw7oe, line 12: remove this as we it's not used anymore. we are using `assertMatch` to assert the attributes instead

instrumentation/opentelemetry_cowboy/src/opentelemetry_cowboy.erl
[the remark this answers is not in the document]
tsloughter: Right. Only set the staus when it is 5xx and remove setting it for codes below that.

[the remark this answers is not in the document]
kw7oe: Okay got it 👍🏻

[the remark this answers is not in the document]
kw7oe: Done on the changes.

tsloughter (approved):

PR #49, opened 2021-12-01
Make Ecto spans of preloads children of the parent span

### LINKED ISSUE #25: Ecto: preloads not linked to parent span
By default, Ecto executes preloads (if there are more than one) in parallel, e.g.,

```elixir
Repo.all(from(Demo.Post, preload: [:authors, :comments]))
```

or 

```elixir
Repo.all(Demo.Post)
|> Repo.preload([:authors, :comments])
```

This will result in the two queries for `authors` and `comments` not to be linked to the parent span.

**Solution**
It would be nice if the preload spans were a child of the parent span (the query for `Demo.Post` in the above example) or that they are created with a link to this parent span (similar to how it's done with Oban jobs in #6).

**Alternatives**
One workaround for this is passing in_parallel: false to Repo.all or Repo.preload, but this might impact the performance.

**Additional context**
Previously reported in https://github.com/opentelemetry-beam/opentelemetry_ecto/issues/15

dvic:
Fixes #25


dvic:
We tried using the published opentelemetry_ecto (where preloads are not linked to the parent span) but it just doesn't work in our situation :) (too many cases where we want to sample spans where preloads are happening). So I decided to migrate the previous PR opentelemetry-beam/opentelemetry_ecto#16 to this repo. I think this PR still stands (or similar changes) if the work mentioned by @bryannaegele in https://github.com/open-telemetry/opentelemetry-erlang-contrib/issues/25#issuecomment-975890127 is done, because something needs to be done in either Ecto itself or the telemetry handlers.

tsloughter:
@bryannaegele this is the stuff you were wanting to move to a dedicated lib, right?

bryannaegele:
@tsloughter yes and I started it a while back. About halfway through. It's the next thing to wrap up on my list after the validation PR as we need it for integrating cowboy with phoenix. Should go quickly.

tsloughter:
@dvic thanks again for this, do you think you could update it to use the new lib that Bryan put together? It isn't merged yet but will be soon: https://github.com/open-telemetry/opentelemetry-erlang-contrib/pull/64

dvic:
@tsloughter given your comment on #64 I made it work (using argument order `depth, key`)

tsloughter:
I don't think the attach/detach strategy used here is correct.

First, the conditioning on context being not undefined seemed odd. If the point is that the parent is wrong if the parent isn't looked up with the process propagator then even if it is undefined wouldn't you want to set it to undefined, since either way the parent will be wrong, may as well have simpler code and remove the 2 if statements?

But also, I think you are duplicating the logic of `with_span` and could instead just use it:

```elixir
parent_context = OpentelemetryProcessPropagator.fetch_parent_ctx(1, :"$callers")
|> Tracer.with_span span_name, ...    
```

bryannaegele:
I tested with the latest propagator and it's working now _but_ the logic here is not quite working as I'd expect. See my note on the nested preloads test.

Only thing I'd call out is it would be _slightly_ more performant to check if there's a current context before using the propagator since the call to local context is checking the local pdict versus always fetching the parent.

instrumentation/opentelemetry_ecto/test/opentelemetry_ecto_test.exs
```diff
@@ -122,6 +125,65 @@ defmodule OpentelemetryEctoTest do
...
+    assert_receive {:span, span(span_id: root_span_id, name: "parent span")}
+    # root query
+    assert_receive {:span, span(parent_span_id: ^root_span_id, name: "opentelemetry_ecto.test_repo.query:users")}
+    # comments preload
+    assert_receive {:span, span(parent_span_id: ^root_span_id, name: "opentelemetry_ecto.test_repo.query:comments")}
+    # users preload
+    assert_receive {:span, span(parent_span_id: ^root_span_id, name: "opentelemetry_ecto.test_repo.query:users")}
+    # preloads of user
+    assert_receive {:span, span(parent_span_id: ^root_span_id, name: "opentelemetry_ecto.test_repo.query:posts")}
+    assert_receive {:span, span(parent_span_id: ^root_span_id, name: "opentelemetry_ecto.test_repo.query:comments")}
```
bryannaegele, line 184: The result here should not be every span having the root span of the overall query as they are nested. If we examine the pdict throughout this test, we can see how the last few queries are spawned by one of the spawned tasks (multiple caller depth).

```
## this query (in a task) spawns the next two, therefore it should be their parent

"#PID<0.434.0> -- SELECT u0.\"id\", u0.\"email\", u0.\"id\" FROM \"users\" AS u0 WHERE (u0.\"id\" = $1)"
{:dictionary,
 [
   "$ancestors": [#PID<0.432.0>],
   ...
   "$callers": [#PID<0.415.0>],
  ...
 ]}

## this is a spawned task from <0.434.0>

21:06:23.496 [debug] QUERY OK source="users" db=0.7ms queue=0.9ms
SELECT u0."id", u0."email", u0."id" FROM "users" AS u0 WHERE (u0."id" = $1) [1]
"#PID<0.437.0> -- SELECT p0.\"id\", p0.\"body\", p0.\"user_id\", p0.\"user_id\" FROM \"posts\" AS p0 WHERE (p0.\"user_id\" = $1) ORDER BY p0.\"user_id\""
{:dictionary,
 [
   "$ancestors": [#PID<0.436.0>],
   "$callers": [#PID<0.434.0>, #PID<0.415.0>],
   "$initial_call": {:erlang, :apply, 2}
 ]}

21:06:23.497 [debug] QUERY OK source="posts" db=0.7ms
SELECT p0."id", p0."body", p0."user_id", p0."user_id" FROM "posts" AS p0 WHERE (p0."user_id" = $1) ORDER BY p0."user_id" [1]

## this is also a spawned task from <0.434.0>

"#PID<0.438.0> -- SELECT c0.\"id\", c0.\"body\", c0.\"user_id\", c0.\"user_id\" FROM \"comments\" AS c0 WHERE (c0.\"user_id\" = $1) ORDER BY c0.\"user_id\""
{:dictionary,
 [
   "$ancestors": [#PID<0.436.0>],
   "$callers": [#PID<0.434.0>, #PID<0.415.0>],
   "$initial_call": {:erlang, :apply, 2}
 ]}

21:06:23.498 [debug] QUERY OK source="comments" db=0.7ms queue=0.7ms
SELECT c0."id", c0."body", c0."user_id", c0."user_id" FROM "comments" AS c0 WHERE (c0."user_id" = $1) ORDER BY c0."user_id" [1]
```

dvic:
> I don't think the attach/detach strategy used here is correct.
> 
> First, the conditioning on context being not undefined seemed odd. If the point is that the parent is wrong if the parent isn't looked up with the process propagator then even if it is undefined wouldn't you want to set it to undefined, since either way the parent will be wrong, may as well have simpler code and remove the 2 if statements?
> 
> But also, I think you are duplicating the logic of `with_span` and could instead just use it:
> 
> ```elixir
> parent_context = OpentelemetryProcessPropagator.fetch_parent_ctx(1, :"$callers")
> |> Tracer.with_span span_name, ...    
> ```

@tsloughter I refactored the code to use this approach, it seems to work just fine 👍 

> I tested with the latest propagator and it's working now _but_ the logic here is not quite working as I'd expect. See my note on the nested preloads test.
> 
> Only thing I'd call out is it would be _slightly_ more performant to check if there's a current context before using the propagator since the call to local context is checking the local pdict versus always fetching the parent.

@bryannaegele I added a check for the current context and I'm going to look into the nested preloads test.

instrumentation/opentelemetry_ecto/test/opentelemetry_ecto_test.exs
dvic: I'm not sure if we can achieve this currently, because I believe Ecto does not emit start/stop events but just a "query" event. So how can we use these spans in between as parent contexts? In the example above, the common ancestor `#PID<0.436.0>` does not have an active span context. And btw, I'm seeing something else when I run the example (just a single entry in `"$callers"` for the preload queries):


```
query: "SELECT u0.\"id\", u0.\"email\" FROM \"users\" AS u0"
#PID<0.415.0>
process info: [
  links: [],
  dictionary: [
    "$__current_otel_ctx": %{
      {:otel_tracer, :span_ctx} => {:span_ctx,
       133037140329346081629534641564206235649, 11340865674359219389, 1, [],
       true, false, true,
       {:otel_span_ets, #Function<2.11792691/1 in :otel_tracer_server.on_end/1>}}
    },
    ...
  ]
]

09:25:41.842 [debug] QUERY OK source="users" db=0.1ms queue=0.2ms
SELECT u0."id", u0."email" FROM "users" AS u0 []
query: "SELECT c0.\"id\", c0.\"body\", c0.\"user_id\", c0.\"user_id\" FROM \"comments\" AS c0 WHERE (c0.\"user_id\" = $1) ORDER BY c0.\"user_id\""
#PID<0.433.0>
process info: [
  links: [#PID<0.431.0>],
  dictionary: [
    "$initial_call": {:erlang, :apply, 2},
    "$ancestors": [#PID<0.431.0>],
    "$callers": [#PID<0.415.0>]
  ]
]
query: "SELECT p0.\"id\", p0.\"body\", p0.\"user_id\", p0.\"user_id\" FROM \"posts\" AS p0 WHERE (p0.\"user_id\" = $1) ORDER BY p0.\"user_id\""
#PID<0.432.0>
process info: [
  links: [#PID<0.431.0>],
  dictionary: [
    "$initial_call": {:erlang, :apply, 2},
    "$ancestors": [#PID<0.431.0>],
    "$callers": [#PID<0.415.0>]
  ]
]
```

bryannaegele: Yeah, it's a good question that I don't have a good answer to 😆

I think we'd need some kind of tracker. I'll have to think on that but what you point out makes sense that it isn't going to be achievable at the moment. This is definitely better than nothing but we should probably call it out to users.

instrumentation/opentelemetry_ecto/lib/opentelemetry_ecto.ex
```diff
@@ -88,22 +88,24 @@ defmodule OpentelemetryEcto do
...
+        current_ctx
+      end
 
-    OpenTelemetry.Span.end_span(s)
+    OpenTelemetry.Tracer.with_span parent_ctx, span_name, %{
+      start_time: start_time,
+      attributes: Map.merge(attributes, base_attributes),
+      kind: :client
+    } do
+      with {:error, error} <- query_result do
```
tsloughter-splunk, line 105: I just realized this isn't doing anything. Why was a span started and ended in this handle event if nothing is happening here? Did you mean to also set the time it ended?

I would not use `with_span` here since you aren't actually wrapping the logic you want to have a span for but I guess are creating a span for something that already happened?

dvic: Yes, the event is emitted once for every query, so indeed, it's creating a span for something that already happened (hence the `start_time`, which is computed from the `total_time` measurement).

instrumentation/opentelemetry_ecto/test/opentelemetry_ecto_test.exs
dvic: Hah okay. I've added some docs on this topic. Feel free to change it to your liking 👍

bryannaegele:
`{:opentelemetry_process_propagator, "~> 0.1.0"}` published

instrumentation/opentelemetry_ecto/lib/opentelemetry_ecto.ex
```diff
@@ -1,6 +1,15 @@
...
+  Telemetry handler for creating OpenTelemetry Spans from Ecto query events. Any
+  relation preloads, which are executed in parallel in separate tasks, will be
+  linked to the span of the process that initiated the call. For example:
+
+      Tracer.with_span "parent span" do
+        Repo.all(Query.from(User, preload: [:posts, :comments]))
+      end
+
+  this will create a span called "parent span" with three child spans for each
+  query: users, posts, and comments.
```
bryannaegele, line 12:
```suggestion
  query: users, posts, and comments.
  
  > #### Note {: .neutral}
  >
  > Due to limitations with how Ecto emits its telemetry, nested preloads are not
  > represented as nested spans within a trace.
```

bryannaegele (commented):
One suggestion on the nested preloads doc.

I'm otherwise good with this. @tsloughter would you prefer reverting the `with_span` back to start/stop?

tsloughter:
Yea, definitely think the `with_span` should be removed. It would be confusing for anyone who might come to this code later I think.

bryannaegele:
@dvic looks great! If you can rebase, we can merge :)

dvic:
> @dvic looks great! If you can rebase, we can merge :)

Done :)

PR #50, opened 2021-12-12
Add the ability to not trace certain routes in opentelemetry_phoenix

henrymori:
Addresses a `ToDo` comment at https://github.com/open-telemetry/opentelemetry-erlang-contrib/blob/3f43f8dd9fc40970448df143121e751ee214b917/instrumentation/opentelemetry_phoenix/lib/opentelemetry_phoenix.ex#L101

Adds the ability to pass in a `config` map with an `ignore_paths` property into `OpentelemetryPhoenix.setup` to ignore certain patjs from being traced e.g. `/metrics`, `/healthz` etc.

Implemented by bypassing the call to `OpentelemetryTelemetry.start_telemetry_span` if the current `request_path` is in the passed in `ignore_paths` map. 

Without:

`OpentelemetryPhoenix.setup()`

With:

`OpentelemetryPhoenix.setup([], %{ignore_paths: ["/healthz", "/metrics"]})`

Particularly important for end-users who may have a large number of nodes sending telemetry data to a SaaS provider that charges per event received.

Open to feedback. Thanks!


bryannaegele:
Hey @henrymori. `sampler` is no longer something that can be set at span creation and not all routes can be determined to be sampled at span start. We unfortunately accidentally left that option in when it shouldn't have been there. These are the final valid start options -> https://github.com/open-telemetry/opentelemetry-erlang/blob/main/apps/opentelemetry_api/src/otel_span.erl#L43-L47

We've discussed the solution for this item will have to be done through creating a proper sampler which is then set on the tracer. The same thing came up with Ecto. One of the open questions has been if the instrumentation libraries should ship samplers with the package for users to opt into which is probably the best route to go.

This would be great to get in, so for direction on a path forward, this functionality needs to go into a sampler and then figure out what a good API is for users of the library.

tsloughter:
At least at this time you just have to not call the start span if it is a certain path, so no relying on a sampler.

henrymori:
Thanks @bryannaegele and @tsloughter for your comments!

Last night I tried for a few hours to get the `sampler` on the `OpenTelemetry.Tracer.with_span` `start_opts` itself to work and got a little lost as I'm not super familiar with the APIs and implementations. I may not be the best person to tackle an implementation in that direction.

I tried the approach suggested by @tsloughter of not starting / stopping a span if the `conn.request_path` is in the config's `ignore_paths` list. I got this implementation to work as it is was conceptually easier to implement, since there are fewer APIs and abstractions for me to wrap my head around :D

Keen to hear your thoughts!

PR #52, opened 2021-12-17
Set phoenix attributes on span creation

matthijsqd:
By setting the attributes inside `OpentelemetryTelemetry.start_telemetry_span/4` instead of afterwards using `OpenTelemetry.Span.set_attributes/2`, they become available in `:otel_sampler` implementations.


bryannaegele (approved):
👏🏻

PR #53, opened 2021-12-28
Add plug dependency for opentelemetry_phoenix

aaronrenner:
I'm currently getting the following compilation warning for opentelemetry_phoenix

    ==> opentelemetry_phoenix
    Compiling 2 files (.ex)
    warning: Plug.Conn.get_peer_data/1 is undefined (module Plug.Conn is not available or is yet to be defined)
      lib/opentelemetry_phoenix.ex:104: OpentelemetryPhoenix.handle_endpoint_start/4

    warning: Plug.Conn.get_req_header/2 is undefined (module Plug.Conn is not available or is yet to be defined)
      lib/opentelemetry_phoenix.ex:211: OpentelemetryPhoenix.header_value/2

Since the opentelemetry_phoenix code is calling functions on `Plug.Conn` directly, the dependency needs to be specified to avoid the compilation warning and ensure the compilation order is correct. 

I also double-checked the plug version requirement and according to the [plug changelog](https://github.com/elixir-plug/plug/blob/v1.6/CHANGELOG.md#v160-2018-06-16) it appears that plug versions as early as 1.6 will work, because that's when `get_peer_data/1` was introduced. However I'd be happy to update the version requirement to a later version if desired.

### DIFF
--- a/instrumentation/opentelemetry_phoenix/mix.exs
+++ b/instrumentation/opentelemetry_phoenix/mix.exs
@@ -61,10 +61,10 @@ defmodule OpentelemetryPhoenix.MixProject do
       {:opentelemetry_api, "~> 1.0.0-rc.4"},
       {:opentelemetry_telemetry, "~> 1.0.0-beta.6"},
       {:telemetry, "~> 0.4 or ~> 1.0.0"},
+      {:plug, ">= 1.11.0"},
       {:cowboy_telemetry, "~> 0.4", only: [:dev, :test]},
       {:opentelemetry, "~> 1.0.0-rc.4", only: [:dev, :test]},
       {:opentelemetry_exporter, "~> 1.0.0-rc.4", only: [:dev, :test]},
-      {:plug, "~> 1.11", only: [:dev, :test]},
       {:ex_doc, "~> 0.26", only: [:dev], runtime: false},
       {:plug_cowboy, "~> 2.4", only: [:dev, :test]},
       {:dialyxir, "~> 1.1", only: [:dev, :test], runtime: false}


bryannaegele:
Hey @aaronrenner, it's a runtime dependency so the compilation warning is fine to ignore. Phoenix pulls it in and we don't really care what version it is. If it bugs ya, we'll want to mark it as optional and keep the version how we had it in test with a `>=` on it.

aaronrenner:
Thanks for the quick review, @bryannaegele! I updated the version requirement to be `>= 1.11` which is what I think you were asking for.

I was also thinking about your request to make the plug dependency optional, but I was thinking if plug isn't installed this wouldn't work because it's explicitly calling functions on `Plug.Conn`. I can still make it optional if you prefer, but I just wanted to double-check. WDYT?

tsloughter:
If it calls Plug directly it should definitely depend on Plug and not rely on it being a transitive dep.

tsloughter (approved):

tsloughter:
I just noticed it is still `~> 1.11` and not `>= 1.11` which seems to be what you two decided on? So I haven't merged.

aaronrenner:
Sorry about that, it must have gotten mixed up when I resolved the merge conflicts. :sweat_smile: 

I personally might consider setting it to `~> 1.10` to match Phoenix 1.5.13 and ensure if Plug 2.x comes out with breaking changes, people won't run into issues. But I'm totally fine with this approach too and I'm just happy to have this dependency specified. :slightly_smiling_face:

bryannaegele:
Earlier versions of Phoenix don't have all the events and data we use, so `>= 1.11` should be fine.

bryannaegele (approved):

PR #64, opened 2022-02-06
Process propagator library

bryannaegele:
We currently don't have any ways to deal with cross-process context propagation which results in broken traces with orphaned spans when instrumenting libraries such as Ecto.

The base functionality to start here is reverse context propagation tools where we can attempt to walk upwards through spawning processes to find an active context. Given the nature of what is being performed here, some APIs will not conform to the propagator spec of `inject` `extract`.

There were no objections to the name when we discussed this at the Observability EG meeting last but I want to wait on adding the CI steps until we're overall happy with this.


dvic:
Have tried it on #49 but I'm getting

```
23:03:15.137 [error] Handler {OpentelemetryEctoTest, #PID<0.539.0>} has failed and has been detached. Class=:error
Reason=:function_clause
Stacktrace=[
  {:lists, :sublist, [[], :"$callers"], [file: 'lists.erl', line: 357]},
  {:opentelemetry_process_propagator, :fetch_parent_ctx, 2,
   [file: 'src/opentelemetry_process_propagator.erl', line: 19]},
...
```

dvic:
Have tried it on #49 but I'm getting

```
23:03:15.137 [error] Handler {OpentelemetryEctoTest, #PID<0.539.0>} has failed and has been detached. Class=:error
Reason=:function_clause
Stacktrace=[
  {:lists, :sublist, [[], :"$callers"], [file: 'lists.erl', line: 357]},
  {:opentelemetry_process_propagator, :fetch_parent_ctx, 2,
   [file: 'src/opentelemetry_process_propagator.erl', line: 19]},
...
```

propagators/opentelemetry_process_propagator/src/opentelemetry_process_propagator.erl
```diff
@@ -0,0 +1,66 @@
...
+         fetch_parent_ctx/0,
+         fetch_parent_ctx/1,
+         fetch_parent_ctx/2]).
+
+-spec fetch_parent_ctx() -> opentelemetry:span_ctx() | undefined.
+fetch_parent_ctx() ->
+    fetch_parent_ctx(1, '$ancestors').
+
+-spec fetch_parent_ctx(non_neg_integer()) -> opentelemetry:span_ctx() | undefined.
+fetch_parent_ctx(MaxDepth) ->
```
tsloughter, line 13: The Elixir API shows this as taking `key` and not depth. The example also passes a key to `fetch_parent_ctx/1` in the Elixir docs. See @dvic 's comment as well.

dvic: I made it work in #49 , however, is it common to have `:undefined` returned in Elixir code? Shouldn't this just be `nil`?

tsloughter: Don't get me started ;). The fact Elixir has `nil` is a big annoyance when doing libs you want used from both Erlang and Elixir.

Since this is Erlang code it is right to be `undefined`, technically the Elixir wrapper could convert it to `nil` but I'm not actually sure this application needs the Elixir wrapper that it comes with. I was going to bring that up with Bryan separately.

I'm fine with it having one, but the Otel API has an Elixir interface because it utilizes macros in both Erlang and Elixir and you can't use Erlang macros in Elixir, so it was required. Then the non-macro wrappers were added to make it feel consistent, since the user would otherwise be forced to use both wrapped and unwrapped which seemed annoying.

bryannaegele: I expect more pieces to this over time where macros would be added. There's also the side benefit of it being much easier to have legit docs. :/

tsloughter: Ah ok.

But docs shouldn't be an issue `opentelemetry` and `opentelemetry_exporter` use `ex_doc` even though they are Erlang apps.

bryannaegele: Bah. I forgot to change the Elixir API.

bryannaegele:
@tsloughter and @dvic - anything else you think I should address? It seems like #49 is working well with this.

dvic:
> @tsloughter and @dvic - anything else you think I should address? It seems like #49 is working well with this.

Nothing from my side, works for me 👍

tsloughter-splunk (approved):

tsloughter (approved):

propagators/opentelemetry_process_propagator/lib/opentelemetry_process_propagator.ex
```diff
@@ -0,0 +1,115 @@
+defmodule OpentelemetryProcessPropagator do
+  @moduledoc """
+  `OpentelemetryProcessPropagator` provides helpers for dealing
+  with context propagation across process boundaries.
+
+  ## Context Propagation
+
+  Erlang and Elxiir do not have a mechanism for transparently passing
```
andrewhr, line 8: typo

```suggestion
  Erlang and Elixir do not have a mechanism for transparently passing
```

propagators/opentelemetry_process_propagator/lib/opentelemetry_process_propagator.ex
```diff
@@ -0,0 +1,115 @@
...
+  @doc """
+  Attempt to fetch an otel context from a give pid.
+  """
+  @spec fetch_ctx(pid) :: OpenTelemetry.span_ctx() | :undefined
+  defdelegate fetch_ctx(pid), to: :opentelemetry_process_propagator
+
+  @doc """
+  Attempt to find an otel context in the spawning process.
+
+  This is equivalent to calling `fetch_parent_ctx(:"$ancestors", 1)`
```
andrewhr, line 93: Given the signature bellow, the example is reversed:

```suggestion
  This is equivalent to calling `fetch_parent_ctx(1, :"$ancestors")`
```

propagators/opentelemetry_process_propagator/src/opentelemetry_process_propagator.erl
andrewhr: 2cc: another idiomatic Elixir idiom is `{:ok, value} | :error`. `Map.fetch/2` does that. Does Erlang uses the same idiom? If it does, that may be an alternative that keep both consistent :)

tsloughter: The reason to return `span_ctx() | undefined` instead of something like `{:ok, value} | :error` is that neither is an error and neither is more correct than the other.

`undefined` is accepted by any function that accepts a SpanCtx (which in most cases just results in nothing being done) because having no SpanCtx is a common case.

In OpenCensus we actually used specific types for this instead of `| undefined` all over, so would be like:

```
-type maybe_span_ctx() :: span_ctx() | undefined.
```

Then functions like this would return `maybe_span_ctx()`.

PR #66, opened 2022-02-14
Add idle_time measurement to opentelemetry_ecto

hoyon:
Ecto exposes a number of time measurements which are already included in the opentelemetry_ecto attributes but `idle_time` was missing.

This PR adds it in.

Ecto docs for it are here: https://hexdocs.pm/ecto/Ecto.Repo.html#module-adapter-specific-events


tsloughter (approved):
Thanks!

PR #68, opened 2022-03-09
Add `http.request_id` attribute

RudolfMan:
- Add `http.request_id` attribute from `resp_headers` set by `Plug.RequestId`.
 - add option to allow configuring the the header name (as it's allowed by `Plug.RequestId`)


RudolfMan:
Actually, I figured, why `request_id` was not included here in first hand.. it's not the part of the standard opentelemetry specs.

tsloughter:
That doesn't mean we can't include it. But it should probably be something like `plug.request_id` or `http.plug.request_id`. I'll have to check if there is a standard for how to define those and whether it should be like `http.<library>.<key>` or not.

bryannaegele:
It is an optional plug, so there would need to be guards around that, as well

PR #69, opened 2022-03-15
elli instrumentation library

tsloughter:


instrumentation/opentelemetry_elli/README.md
```diff
@@ -0,0 +1,91 @@
...
+> http.server_name has shown great value in practice, as bogus HTTP Host headers occur often in the wild. It is strongly recommended to set http.server_name to allow associating requests with some logical server entity.
+
+## Use
+
+### Including the Middleware and setting Span names
+
+The middleware takes care of extracting the parent Span from the requests
+headers, both the [W3C](https://w3c.github.io/trace-context/) and [B3](https://github.com/openzipkin/b3-propagation) formats are supported.
+
+Because Elli has no router there is not a way to get a very descriptive Span
```
ferd, line 40:
```suggestion
Because Elli has no router, there is no way to get a very descriptive Span
```

instrumentation/opentelemetry_elli/README.md
```diff
@@ -0,0 +1,91 @@
...
+The macro `update_name` from `opentelemetry_api/include/otel_tracer.hrl` allows you
+to update the name of the Span after it has started:
+
+``` erlang
+handle(Req, Args) ->
+    handle(Req#req.path, Req, Args).
+
+handle([<<"hello">>, Who], Req, _Args) ->
+    ?update_name(<<"/hello/{who}">>),
+    {ok, [], <<"Hello ", Who/binary>>}.
```
ferd, line 54: This could possibly be done with a parse transform, but would require specifying a function/arity to always respect. Not required at this point for sure.

instrumentation/opentelemetry_elli/README.md
```diff
@@ -0,0 +1,91 @@
...
+```
+
+Attributes set by the middleware can be found in the OpenTelemetry docs [Semantic
+conventions for HTTP spans](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/data-http.md).
+
+Note that it isn't required to update the Span name, that is just a suggestion
+for when it makes sense to do so. And another option is to create a child Span
+with this better name and its parent will be the Span named `"HTTP
+{METHOD_NAME}"`.
```
ferd, line 63:
```suggestion
Updating the Span name is not required. Another option is to create a child Span
with a more descriptive name, whose parent will be the Span named `"HTTP
{METHOD_NAME}"`.
```

instrumentation/opentelemetry_elli/README.md
```diff
@@ -0,0 +1,91 @@
...
+
+### Excluding Paths
+
+Not all paths are created equal. It is likely you don't want to Trace a health
+check endpoint or `/metrics` if using Prometheus to scrape metrics. So this
+library offers two ways to exclude requests from potentially creating Spans by
+filtering on the raw path in the URL.
```
ferd, line 70:
```suggestion
check endpoint or `/metrics` if using Prometheus to scrape metrics. This
library offers two ways to exclude requests from potentially creating Spans by
filtering on the raw path in the URL.
```

instrumentation/opentelemetry_elli/README.md
```diff
@@ -0,0 +1,91 @@
...
+
+A Application environment variable is read in on Elli start, so the follow can
```
ferd, line 72:
```suggestion
An application environment variable is read in on Elli start, so the following can
```

instrumentation/opentelemetry_elli/README.md
```diff
@@ -0,0 +1,91 @@
...
+be added to `sys.config` to exclude the url `/exclude/me`:
```
ferd, line 73:
```suggestion
be added to `sys.config` to exclude the URLs `/health` and `/metrics`:
```

instrumentation/opentelemetry_elli/README.md
```diff
@@ -0,0 +1,91 @@
...
+
+``` erlang
+
+{opentelemetry_elli, [{excluded_urls, ["/health", "/metrics"]}]}
+```
+
+An OS environment variable, `OTEL_ELLI_EXCLUDED_URLS`, is also read and is
+a comma separated list of paths.
+
+The lists from both are merged.
```
ferd, line 83:
```suggestion
The OS environment variable `OTEL_ELLI_EXCLUDED_URLS` is also read and will be merged
with the `sys.config` value. It should be a comma separated list of paths.
```

instrumentation/opentelemetry_elli/src/opentelemetry_elli.app.src
```diff
@@ -0,0 +1,15 @@
+{application, opentelemetry_elli,
+ [{description, "Elli middleware for tracing and stats with OpenTelemetry"},
+  {vsn, "git"},
+  {registered, []},
+  {applications,
+   [kernel,
+    stdlib,
+    opentelemetry_api
+   ]},
```
ferd, line 9:
```suggestion
  {applications,
   [kernel,
    stdlib,
    opentelemetry_api,
    elli
   ]},
```

instrumentation/opentelemetry_elli/README.md
ferd: also can the OTEL_ELLI_EXCLUDED_URLS be quoted, contain commas, be escaped?

ferd (approved):

instrumentation/opentelemetry_elli/README.md
tsloughter: Do you mean a pt that is able to take `[<<"hello">>, Who]` and create `<<"/hello/{Who}">>`? I like that. 

But what do you mean about specifying function/arity?

instrumentation/opentelemetry_elli/README.md
tsloughter: Good point, adding a note that it is simply done with a string `split` at this time, so commas in the url would break it.

tsloughter: Done. Took more words than I wanted since I wanted to explain you can still percent encode and that this is how the URL would be seen by Elli when comparing them anyway.

instrumentation/opentelemetry_elli/README.md
ferd: yeah that's what I mean. The function arity would be that you go from handle/2 (which is probably registered or standard somewhere, I forget) to handle/3, so it's not guaranteed that handle/3 would be the callback that needs transforming.

tsloughter: Ah, yea, if it were done I'd just define it as `handle/3`, requiring the user do the call from `handle/2` if they want to use it.

PR #77, opened 2022-04-11
Relax requirement on telemetry library

GregMefford:
I noticed that these `telemetry` version restrictions are probably more strict than they should be, because `telemetry` `1.1.0` has been released, and it's showing as not possible to upgrade because of these version specs limiting it to `1.0.*`.

I didn't make updates in the Rebar configs because I'm not familiar with how that works. I just updated the Mix configs.


tsloughter-splunk (approved):

tsloughter (approved):

tsloughter:
Looks like all the rebar ones are already `~> 1.0`?

GregMefford:
FYI I don't have merge permissions, in case you were waiting on me to press the button. 😅

bryannaegele (approved):

PR #81, opened 2022-04-29
Report Redix error messages through OTel status

andrewhr:
Instead of custom attributes, leverage the status description as described in Semantic Conventions. This approach is taken from current `opentelemetry_ecto` implementation.

Small non-related change is a fix the license description in `mix.exs`.

### DIFF
--- a/instrumentation/opentelemetry_redix/lib/opentelemetry_redix.ex
+++ b/instrumentation/opentelemetry_redix/lib/opentelemetry_redix.ex
@@ -60,7 +60,6 @@ defmodule OpentelemetryRedix do
       }
       |> Map.merge(net_attributes(connection))
       |> Map.merge(redix_attributes(meta))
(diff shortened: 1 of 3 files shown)


instrumentation/opentelemetry_redix/lib/opentelemetry_redix.ex
```diff
@@ -69,8 +68,12 @@ defmodule OpentelemetryRedix do
         attributes: attributes
       })
 
-    if meta[:reason] do
-      OpenTelemetry.Span.set_status(s, OpenTelemetry.status(:error, ""))
+    case meta do
```
bryannaegele, line 71: This could still be an `if` statement since nothing is happening in the other case.

`if meta[:kind] == :error do`

bryannaegele (approved):

PR #82, opened 2022-05-06
Fix two use cases of Ecto span propagation

andrewhr:
First one is related to `OpenTelemetry.Ctx` API.

I've noticed in a few scenarios the current span of a trace may get lost after Ecto calls. Looking at the The `attach/1` typespec, it's a Ctx -> Token, while `dettach/1` as Token -> Ctx function. That made me assume the expected input of dettach is the return type of attach. Indeed, after this change we got the behavior of Ecto calls preserve the parent span untouched.

That leads to a second bug found. When ecto does simple calls within a Task, due the special propagation code for preloads that means it will skip the current span, if any.  The solution here is to first check the current process.

One test was added to reproduce this bug.

### DIFF
--- a/instrumentation/opentelemetry_ecto/lib/opentelemetry_ecto.ex
+++ b/instrumentation/opentelemetry_ecto/lib/opentelemetry_ecto.ex
@@ -108,11 +108,21 @@ defmodule OpentelemetryEcto do
       |> Map.merge(base_attributes)
       |> Map.merge(additional_attributes)
 
-    parent_context = OpentelemetryProcessPropagator.fetch_parent_ctx(1, :"$callers")
+    parent_context =
+      case OpentelemetryProcessPropagator.fetch_ctx(self()) do
+        :undefined ->
+          OpentelemetryProcessPropagator.fetch_parent_ctx(1, :"$callers")
 
-    if parent_context != :undefined do
-      OpenTelemetry.Ctx.attach(parent_context)
-    end
+        ctx ->
+          ctx
+      end
+
+    parent_token =
+      if parent_context != :undefined do
+        OpenTelemetry.Ctx.attach(parent_context)
+      else
+        :undefined
+      end
 
     s =
       OpenTelemetry.Tracer.start_span(span_name, %{
@@ -131,8 +141,8 @@ defmodule OpentelemetryEcto do
 
     OpenTelemetry.Span.end_span(s)
 
-    if parent_context != :undefined do
-      OpenTelemetry.Ctx.detach(parent_context)
+    if parent_token != :undefined do
+      OpenTelemetry.Ctx.detach(parent_token)
     end
   end
 
--- a/instrumentation/opentelemetry_ecto/test/opentelemetry_ecto_test.exs
+++ b/instrumentation/opentelemetry_ecto/test/opentelemetry_ecto_test.exs
@@ -192,6 +192,29 @@ defmodule OpentelemetryEctoTest do
     assert_receive {:span, span(parent_span_id: ^root_span_id, name: "opentelemetry_ecto.test_repo.query:comments")}
   end
 
+  test "nested query within Task does not skip parent span" do
+    user = Repo.insert!(%User{email: "opentelemetry@erlang.org"})
+    Repo.insert!(%Post{body: "We got traced!", user: user})
+    Repo.insert!(%Comment{body: "We got traced!", user: user})
+
+    attach_handler()
+
+    Tracer.with_span "root span" do
+      task =
+        Task.async(fn ->
+          Tracer.with_span "parent span" do
+            Repo.all(User)
+          end
+        end)
+
+      Task.await(task)
+    end
+
+    assert_receive {:span, span(span_id: _root_span_id, name: "root span")}
+    assert_receive {:span, span(span_id: parent_span_id, name: "parent span")}
+    assert_receive {:span, span(parent_span_id: ^parent_span_id, name: "opentelemetry_ecto.test_repo.query:users")}
+  end
+
   def attach_handler(config \\ []) do
     # For now setup the handler manually in each test
     handler = {__MODULE__, self()}
--- a/instrumentation/opentelemetry_ecto/test/test_helper.exs
+++ b/instrumentation/opentelemetry_ecto/test/test_helper.exs
@@ -1,5 +1,5 @@
 OpentelemetryEcto.TestRepo.start_link()
 
-ExUnit.start()
+ExUnit.start(capture_log: true)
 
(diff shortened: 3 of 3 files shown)


bryannaegele:
Can you provide a top-level ecto query where this would occur in the test? I'm unaware of a case where ecto uses tasks for anything except preloads.

instrumentation/opentelemetry_ecto/test/opentelemetry_ecto_test.exs
```diff
@@ -184,6 +184,29 @@ defmodule OpentelemetryEctoTest do
...
   end
 
+  test "nested query within Task does not skip parent span" do
+    user = Repo.insert!(%User{email: "opentelemetry@erlang.org"})
+    Repo.insert!(%Post{body: "We got traced!", user: user})
+    Repo.insert!(%Comment{body: "We got traced!", user: user})
+
+    attach_handler()
+
+    Tracer.with_span "root span" do
```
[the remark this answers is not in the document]
bryannaegele: Ok, so this is solely in the context of another Task. Got it.

instrumentation/opentelemetry_ecto/lib/opentelemetry_ecto.ex
```diff
@@ -102,11 +102,21 @@ defmodule OpentelemetryEcto do
           acc
       end)
 
-    parent_context = OpentelemetryProcessPropagator.fetch_parent_ctx(1, :"$callers")
+    parent_context =
+      case OpentelemetryProcessPropagator.fetch_ctx(self()) do
+        :undefined ->
```
bryannaegele, line 107: This should just attempt to get the context directly and not go through the propagator.

```suggestion
      case OpenTelemetry.Ctx.get_current() do
        %{} ->
```

instrumentation/opentelemetry_ecto/lib/opentelemetry_ecto.ex
```diff
@@ -125,8 +135,8 @@ defmodule OpentelemetryEcto do
 
     OpenTelemetry.Span.end_span(s)
 
-    if parent_context != :undefined do
-      OpenTelemetry.Ctx.detach(parent_context)
+    if parent_token != :undefined do
+      OpenTelemetry.Ctx.detach(parent_token)
```
bryannaegele, line 139: The token and context are both maps and they're both the context.

bryannaegele (commented):
Before you address any of these comments, I'm going to change the process propagator to return `#{}` instead of `undefined` to match the behavior of `otel_ctx`. That should make handling more consistent.

instrumentation/opentelemetry_ecto/lib/opentelemetry_ecto.ex
andrewhr: Empty map matches any map, so this pattern can't distinguish between "no current ctx" from "any ctx". I could change to something like:

```suggestion
      case OpenTelemetry.Ctx.get_current() do
        ctx when map_size(ctx) == 0 ->
```

The approach of `OpentelemetryProcessPropagator.fetch_ctx` is just so the code that manipulate the ctx here is consistent (everything goes through the propagator). I can change the strategy to the proposed one if you prefer :)

bryannaegele: Maybe we should reference current span ctx instead since it will return undefined if there isn't a span in progress.

https://github.com/open-telemetry/opentelemetry-erlang/blob/main/apps/opentelemetry_api/src/otel_tracer.erl#L121

@tsloughter any thoughts here?

tsloughter: what is it trying to do here?

instrumentation/opentelemetry_ecto/lib/opentelemetry_ecto.ex
tsloughter: Still better to use the name "token" and pretend it isn't a context just to not rely on it in case something were to change in the future. It probably won't... but still

instrumentation/opentelemetry_ecto/lib/opentelemetry_ecto.ex
andrewhr: This instrumentation may run from inside Task. There is the case of preloads which spawn a Task and we want to find the context from $callers, to connect spans. If the query is not a preload, there is no intermediary Task.

In case of no intermediary Task (from preload), this code will skip a legit Task from client code. The solution is to search the current process first, to have an opportunity to short-circuit the problematic scenario.

tsloughter: aah. But if its not overriding the context and uses the macros to start/end spans then it should already use any current context that might exist in the process. There should be no need to get the current context to attach it as far as I can understand it.

bryannaegele:
@andrewhr check out #86 and let me know if the new function I added simplifies things.

andrewhr:
@bryannaegele I did a try locally and I was able to reduce the amount of changes so I say 👍

jmbutter:
I ran into these same span propagation issues while instrumenting some custom background workers that heavily rely on Tasks.  The main problem being that after running an Ecto query in one of those tasks, any further calls to `OpenTelemetry.Tracer.set_attributes` would place the attributes on the parent span instead of the span that was created inside the task.  See [here](https://github.com/open-telemetry/opentelemetry-erlang-contrib/pull/98/files#diff-9de6a7bf4c268ec80d75dadb15b366d3a187f698590b214d44bd21674d242128R217) for an example test case that illustrates it, from my attempt at solving this issue before I realized this fix was already in progress.

At any rate, I have tested this solution and it fixes all of my issues.  Would it be possible to get this merged while the other changes discussed are ironed out?

tsloughter:
Is this ready to merge and can just be updated later for @bryannaegele's changes to the other library?

tsloughter:
oh, hehe, @jmbutter asked the same.

We can also discuss in the SIG meeting today if anyone can make it.

bryannaegele:
@andrewhr could you revert the changes from using the propagator to the previous commit you had that had implemented that functionality?

andrewhr:
@bryannaegele hey, I did some code spelunking here and realized the code that uses Propagator changes has naver been pushed, so the code you saw here is the original change - the exising Propagator calls are (almost) the same we have on main.

So, I just rebased the code to ensure it's mergeable.

lshi18:
Hi, it would be nice to learn the state and the plans of this PR. Thanks.

bryannaegele (approved):

PR #86, opened 2022-05-24
Match otel ctx returns and add get_ctx helper

bryannaegele:
Modifying the return values from `undefined` to an empty map to match the behavior of `otel_ctx` when a context is not found.

I've also added a convenience function `get_ctx` which matches the behavior of `fetch_parent_ctx` functions but first attempts checks the current process before attempting to search parents. Hopefully this simplifies needing to perform multiple ctx checks when users can't be sure if the code is being invoked within a spawned process or an already active span - looking at you, Ecto.


tsloughter:
As discussed in the SIG meeting: it is likely better to rename `fetch_parent_ctx` to `extract` and instead of returning the context as-is first grab the span ctx and use `otel_tracer:from_remote_span` to create a span ctx to put in the extracted context instead.

This makes the propagator act more like the other propagators, even if it doesn't have an injection (at this time), since it must work in cases like Ecto where it isn't possible to call an injection since they are instrumented with otel -- and so extraction uses the ability to grab the pdict of another process to find its parent.

I think it should still do all this even if it is returning what is actually a active span in the same process. It is a bit weird, but my understanding from the Ecto use case is that this is needed.

PR #88, opened 2022-05-27
Telemetry-based Absinthe instrumentation

andrewhr:
Add instrumentation for `Absinthe`, the GraphQL Elixir-framwork. Relies on some new telemetry work on Absinthe:
* [Added telemetry for async helper][1], used to propagate traces
* [Improved telemetry for batch helper][2], used to propagate traces
* [Tweaked pipeline instrumentation][3], that splits parse/validate/execute
  phases, following the approach of [JavaScript libraries][4]

Span names and attributes also follow the [JavaScript impl][4], the closes to a future [Semantic Attribute][5] spec.

The "span tracker" used here is to connect traces in a shape that resembles the query shape instead of execution shape. One example is:

```
query {
  users {
    name
  }
}
```

In case `name` is an async resolver (for example), the execution will resembles a recursive function, and such our trace will be:

```
RootQueryType.users
  User.name
    User.name
      User.name
```

With the span tracker, the above trace will become:

```
RootQueryType.users
  User.name
  User.name
  User.name
```

That helps visually scan the traces, and it's friendly to folks who doesn't know much about Absinthe internals.

[1]: https://github.com/absinthe-graphql/absinthe/pull/1169
[2]: https://github.com/absinthe-graphql/absinthe/pull/1170
[3]: https://github.com/absinthe-graphql/absinthe/pull/1172
[4]: https://www.npmjs.com/package/@opentelemetry/instrumentation-graphql
[5]: https://github.com/open-telemetry/opentelemetry-specification/pull/2456


rewritten:
Hey @andrewhr could you make the absinthe WIP features you are relying on more explicit? I'm considering pulling in the code (until the PR can be merged and the package adopted from @primait) in our current project.

andrewhr:
Hey @rewritten!

I linked all Absinthe issues in the description of this issue here.

The gist is, we need telemetry events for all kinds of helpers/middleware (async/batch), and they need to be fired from inside the internal Tasks created by those middleware (so span propagation works, as it tracks spans from parent processes).

There is one case which events exists, but they're not well located and propagation does work - this specific case is handled  here https://github.com/absinthe-graphql/absinthe/pull/1170.

tsloughter:
@andrewhr coming up on a year so I thought I'd check on if there was anything knew around absinthe integration?

PR #91, opened 2022-06-04
Record exception for ecto_query error

spencerdcarlson:
Record the exception for ecto query errors.


spencerdcarlson:
> are all error results in the telemetry event exceptions?
I think passing it through `Exception.normalize/2` will take care of any cases where it is not an exception.

```elixir
iex(11)> Exception.normalize(:error, %UndefinedFunctionError{})
%UndefinedFunctionError{
  arity: nil,
  function: nil,
  message: nil,
  module: nil,
  reason: nil
}
iex(12)> Exception.normalize(:error, nil)
%ErlangError{original: nil}
iex(13)> Exception.normalize(:error, "some string")
%ErlangError{original: "some string"}
```

bryannaegele:
Yes, but your code isn't accounting for if it is an exception or not, it just records an exception no matter what in the current implementation. How can we distinguish between those cases?

spencerdcarlson:
> Yes, but your code isn't accounting for if it is an exception or not, it just records an exception no matter what in the current implementation. How can we distinguish between those cases?

I guess I don't follow your question. My understanding is that this will only take place when ecto_sql tells us that that query returned an error `{:error, error}` — Wouldn't we always want to capture the error details regardless of if is a formal Exception or not?

Do you think we need some extra filtering like `OpentelemetryEcto.format_error/1` to ensure that just because ecto says the query returned an error we only capture error details if it is a formal exception?

spencerdcarlson:
> Yes, but your code isn't accounting for if it is an exception or not, it just records an exception no matter what in the current implementation. How can we distinguish between those cases?

Ok, I think I follow. Is the concern that if ecto returns. `{:error, "Something bad happened"}` then we are essentially type casting that string into an `ErlangError`. When it isn't actually an exception? My thought is that we would still want to capture any details we can about the error and propagate that as an OpenTel exception.

bryannaegele:
Yeah, the point was we can't assume everything is an exception just because the adapter returned an error and we can only record something as an exception if it actually is. We already set an error and capture the error message so I think any change here would have to be additive only for real exceptions. Since it's up to the adapter to emit and populate this event, I'm not sure if any of them are going to return an exception struct as the second value in the error tuple.

tsloughter:
Right, if it isn't an exception then the error on the status should be enough.

spencerdcarlson:
> I'm not sure if any of them are going to return an exception struct as the second value in the error tuple.

You can see from the augmented test that it is already returning an ` %Postgrex.Error{}` as the second tuple. Also we are throwing away the error message if it is not an exception. What do you think about something like this?

```elixir
...
  case query_result do
      {:error, %{__exception__: true} = exception} ->
        OpenTelemetry.Span.record_exception(s, Exception.normalize(:error, exception))
        OpenTelemetry.Span.set_status(s, OpenTelemetry.status(:error, Exception.message(exception)))

      {:error, error} ->
        status = if utf8?(error), do: error, else: ""
        OpenTelemetry.Span.set_status(s, OpenTelemetry.status(:error, status))

      {:ok, _} ->
        :ok
    end

    OpenTelemetry.Span.end_span(s)

    if parent_context != :undefined do
      OpenTelemetry.Ctx.detach(parent_context)
    end
  end

defp utf8?(<<_::utf8, rest::binary>>), do: utf8?(rest)
defp utf8?(<<>>), do: true
defp utf8?(_), do: false
```  

The reason I am opening this MR is because an exception gets surfaced in data dog and an error message does not. It seems that [opentelemetry_phoenix](https://github.com/open-telemetry/opentelemetry-erlang-contrib/blob/main/instrumentation/opentelemetry_phoenix/lib/opentelemetry_phoenix.ex#L161-L184) is instrumented to provide those details

tsloughter:
@spencerdcarlson to be clear, does that above code that matches out an exception case work for your use-case and you see the error in datadog? 

I think it is reasonable to record the exception if it is an exception -- like in your example code -- but it'd be good to know also why datadog isn't surfacing the error in `OpenTelemetry.Span.set_status(s, OpenTelemetry.status(:error, format_error(error)))`

tsloughter:
I figured I'd see what others were doing and Python sqlalchemy is recording errors like this the same way: https://github.com/open-telemetry/opentelemetry-python-contrib/blob/3137c1fa413a2b530300bfed57f71dcc94412b6a/tests/opentelemetry-docker-tests/tests/sqlalchemy_tests/test_sqlite.py#L59

It maybe should be added to the database semantic conventions though.

tsloughter:
I assume you are using the otel collector with the datadog exporter?

I dug around in there a bit, has some code that looks specific to figuring out error information from the span status message https://github.com/open-telemetry/opentelemetry-collector-contrib/commit/a95cdc07ea0fe6551604053015916dc680e35937

spencerdcarlson:
> @spencerdcarlson to be clear, does that above code that matches out an exception case work for your use-case and you see the error in datadog?

Yes, I believe so. The exceptions that are emitted by the [opentelemetry_phoenix](https://github.com/open-telemetry/opentelemetry-erlang-contrib/blob/main/instrumentation/opentelemetry_phoenix/lib/opentelemetry_phoenix.ex#L161-L184) have great visibility in data dog. The main difference that I notice is the call to [OpenTelemetry.Span.record_exception/4](https://github.com/open-telemetry/opentelemetry-erlang-contrib/blob/main/instrumentation/opentelemetry_phoenix/lib/opentelemetry_phoenix.ex#L178). In the email I sent you there are two screenshots of a DB error trace and an HTTP error trace and you can see the difference. I don't want to include the screenshots here since they have some internal code references.

> but it'd be good to know also why datadog isn't surfacing the error in OpenTelemetry.Span.set_status(s, OpenTelemetry.status(:error, format_error(error)))

I agree. This seems like a bug that should be opened with Data Dog. 

I will update this branch with my proposed solution unless either of you have any objections.

tsloughter:
Interesting, so that means it is indeed an exception. It is beside the point I'm just thrown off by Ecto returning exceptions as errors.

Do you have anyone at Datadog to also raise this issue with? I can try to poke around to find someone.

spencerdcarlson:
> Interesting, so that means it is indeed an exception. It is beside the point I'm just thrown off by Ecto returning exceptions as errors.

That's the default behavior of Ecto/[[Postgrex](https://hexdocs.pm/postgrex/Postgrex.html#execute/4)|[MyXQL](https://hexdocs.pm/myxql/MyXQL.html#execute/4)]
```elixir
iex(1)> Repo.query("SELECT * FROM foo")
[debug] QUERY ERROR db=2.3ms queue=8.6ms idle=225.2ms
SELECT * FROM foo []
{:error,
 %Postgrex.Error{
   connection_id: 19802,
   message: nil,
   postgres: %{
     code: :undefined_table,
     file: "parse_relation.c",
     line: "1180",
     message: "relation \"foo\" does not exist",
     pg_code: "42P01",
     position: "15",
     routine: "parserOpenTable",
     severity: "ERROR",
     unknown: "ERROR"
   },
   query: "SELECT * FROM foo"
 }}
 ```

> Do you have anyone at Datadog to also raise this issue with? I can try to poke around to find someone.

I will keep looking into why Datadog or the OTEL collector is not surfacing the error message.

Do you have any reservations about going forward with the proposed code change in the meantime?

**Comment by tsloughter** [comment]
Not positive yet. But that stacktrace doesn't look right. It is a trace to the telemetry library, not where the query was called from.

**Comment by tsloughter** [comment]
@mx-psi thanks!

I'm not sure there is an issue with the exporter. I was noting that there is code in the exporter specific to handling the status error message, so it was possible that we might find an explanation there. But I hadn't noticed anything that stands out as a potential cause.

**Comment by spencerdcarlson** [comment]
@mx-psi here are some redacted (just being overly cautions) screenshots of how I see the issue. 

All the error traces related to `opentelemetry_ecto` only display `error: true`. The error status message has the helpful details (e.g. `"[ERROR 42703 (undefined_column) column u0.non_existant_field does not exist\n\n query: SELECT u0."non_existant_field" FROM "users" AS u0]"`) but it is not associated to the trace.

In contrast it appears that most `opentelemetry_phoenix` error events contain a stacktrace and even an `error.message` tag. (e.g. `"expected at most one result but got 143 in query: from i0 in SelfServiceIntegration.Integration.Schema.Integration, preload: [[:tags, :tray_solution]]"`)

**opentelemetry_ecto Error Trace**
opentelemetry_ecto-error-trace-redacted

**opentelemetry_phoenix Error Trace**
opentelemetry_phoenix-error-trace-redacted

**Comment by spencerdcarlson** [comment]
> What do you mean by "but it is not associated to the trace."? That information is somewhere in the Datadog UI but just not shown in the span tags?

I can not find it anywhere in Datadog. It appears to not be in Datadog, but I don't know exactly where the error status is getting lost.

**Comment by spencerdcarlson** [comment]
> Not positive yet. But that stacktrace doesn't look right. It is a trace to the telemetry library, not where the query was called from.

This stack trace points directly back to where the original query was executed. Notice the line : `test/opentelemetry_ecto_test.exs:114` which is referencing `Repo.all(from u in "users", select: u.non_existant_field)` from the test. 

The reason that the first two lines in the stacktrace have telemetry in it is because after Ecto executes the query, it calls telemetry to emit the event -- so it makes sense that telemetry calls are in the stacktrace. 

```
App -> ecto -> db_connection -> ecto_sql -> telemetry -> App.event_handler
```

This thread is attacking two distinct issues:
1. The error status message is getting lost. (@mx-psi is looking into it)
1.  Not associating exceptions to a given trace.

Can we agree to fix 2. by reporting the exception?

**Comment by tsloughter** [comment]
@spencerdcarlson what I see is the stacktrace starts with `(opentelemetry_ecto 1.0.0) lib/opentelemetry_ecto.ex:120: OpentelemetryEcto.handle_event/4` because it isn't a stacktrace to where the exception happened, but to where the library is recording the exception.

I didn't see where the stacktrace was coming from since I don't see it in the PostgrexError, so I looked up the format function and it says it will use the process info to get the current stacktrace: https://hexdocs.pm/elixir/1.12/Exception.html#format_stacktrace/1

That would explain `handle_event` being the top function of the stack.

You have other examples that are exceptions, like the one that times out in the transaction and has a stacktrace to the call that was running when it timed out.

If all the information is available in the Status then I don't think we should add an exception, unless there is an exception that must be caught and thus also has a stacktrace.

**Comment by tsloughter** [comment]
Note that the Phoenix integration only records an exception if the exception event is triggered, which includes a stacktrace.

**Comment by bryannaegele** [comment]
There's a few more nuances here that make this difficult that we've discussed since the beginning of this lib.

1. Are we expecting this library to eventually translate and implement deciphering error vs returned exception for all adapters in the ecosystem? We'd really like the other libraries and drivers to have their own instrumentation, e.g. Cowboy is instrumented separately from Phoenix.
  a. Each of the built in adapters have different values on their exceptions. [Postgres 1](https://hexdocs.pm/postgrex/0.15.5/Postgrex.Error.html#content), [Postgres 2](https://hexdocs.pm/postgrex/0.15.5/Postgrex.QueryError.html#content), [MyXQL](https://hexdocs.pm/myxql/0.6.1/MyXQL.Error.html#summary), TDS has none, SQLite has none, etc.
2. These telemetry fields events are:
  a. Completely optional
  b. Left to each adapter to implement


I do think there is a clear issue with DD if it isn't marking these spans as errored which, while frustrating, isn't going to influence any decision to mark everything an exception without a holistic review for all of Ecto's use cases. They show up fine in Lightstep and other tooling, so that seems like a bug on their end.

**Comment by spencerdcarlson** [comment]
@tsloughter sorry, where is [Exception.format_stacktrace/1](https://hexdocs.pm/elixir/1.12/Exception.html#format_stacktrace/1) being called? 

I believe the stacktrace is coming `%Postgrex.Error{}` and I don't think that the call to [Exception.normalize/2](https://github.com/elixir-lang/elixir/blob/v1.13.4/lib/elixir/lib/exception.ex#L98) is even needed because if `%{__exception__: true} = exception` then it just passes through the exception. This was protection for a possible scenario in my fist commit that we can pull out.

I think the `handle_event` is in the stack trace because telemetry uses dynamic dispatching to emit the metric so the call to record the metric happens inside `ecto_sql` but in the stacktrace of the process that originally executed the query. 

ecto_sql v3.7.2 [emit event](https://github.com/elixir-ecto/ecto_sql/blob/v3.7.2/lib/ecto/adapters/sql.ex#L927) via telemetry
Telemetry - [emit event](https://github.com/beam-telemetry/telemetry/blob/main/src/telemetry.erl#L154-L177) via dynamic dispatching.

Notice how v3.7.3 (current mix.lock version for opentelemetry_ecto) its Line L927 which is in the stacktrace as `(ecto_sql 3.7.2) lib/ecto/adapters/sql.ex:927: Ecto.Adapters.SQL.log/4`

**Comment by bryannaegele** [comment]
> Note that the Phoenix integration only records an exception if the exception event is triggered, which includes a stacktrace.

Right. And for reference, the spec is what we have to work from. [`record_exception` is just a convenience for `add_event`](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/api.md#record-exception) with some formatting niceties. In the case of an exception, you have to use one way to record the event or the other, but not both.

**Comment by tsloughter** [comment]
@spencerdcarlson I don't see stacktrace in the postgrex error struct. `format_stacktrace` is called by `record_exception`. `handle_event` line 120 is the top of the stack because it is getting the current stacktrace where you are calling `record_exception` and not a stacktrace to where the error occured.

**Comment by spencerdcarlson** [comment]
@tsloughter thanks for the catch!

**Comment by gbbr** [comment]
Hey there 👋🏻 I work on APM, and the `datadogexporter`. It is not entirely clear to me what the issue is from reading this. There is a lot of information sprinkled around the responses which is specific to another project so it's hard for me to make sense of it.

Can you please help me understand the specific issue that exists inside the `datadogexporter`? Perhaps we can open a new issue specific to that in https://github.com/open-telemetry/opentelemetry-collector-contrib/issues ?

It looks to me that if the status code of the trace is set to error, and if the `error.msg` tag is not specifically set by the user, it will be set to the status message [here](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/ac658f3c9f07981eff77f9397931932061670f2e/exporter/datadogexporter/translate_traces.go#L596).

In order to help, it'd be helpful to me if I could understand the issue specifically related to the context of the exporter.

File: instrumentation/opentelemetry_ecto/test/opentelemetry_ecto_test.exs [hunk 1]
```diff
@@ -111,18 +111,47 @@ defmodule OpentelemetryEctoTest do
     attach_handler()
 
     try do
-      Repo.all(from u in "users", select: u.non_existant_field)
+      Repo.all((from u in "users", select: u.non_existant_field), [stacktrace: true])
```
spencerdcarlson, line 114: We can set this in `instrumentation/opentelemetry_ecto/config/test.exs` to be a bit more clean.

tsloughter:
So the stacktrace is included if the user adds a `:stacktrace` option on each query?

Maybe if the PR took into account whether it was an "accurate" stacktrace somehow and only included it as an exception in that case I could see it being added. Would that be possible?

Since the error status can not have a stacktrace I see the desire for having an exception event, but if the stacktrace isn't guaranteed to be to the right place then we should only record the exception when it is.

tsloughter:
Actually, is the stacktrace actually useful in this scenario? Particularly if it is a failure of the query I wouldn't think so. It would simply point you to the query, which is already given in the attributes.

The information in red here https://user-images.githubusercontent.com/1206303/172432438-99dacd7a-5dec-478c-8ef2-738a7c72395a.png should be what the status is already (minus the query since that is a separate attribute of the span.

yordis (approved):

PR #92, opened 2022-06-09
Add Broadway instrumentation library

andrewhr:
Library to instrument Broadway topologies, including trace linking.

Broadway is a tricky one to describe in OpenTelemetry. It feels like a good one to follow Messaging semantics (like Oban), though they're doesn't plainly map to it.

The approach taken here are usage of custom attributes to have a good start, while waiting for those Messaging semantics to stabilize - looks like they will change a lot anyway [in some near future][1].

One alternative is to model each GenStage as it's own destination, sort-of.

Another experiment done in this patch is adding some means to propagate traces as links. Broadway doesn't give great means to do without some friction. The approach here is to add a wrapper for `Broadway.start_link/2` to inject extra configuration that will take care of this context extraction and propagation.

It relies on Producer [`transformers`][2] and peek on some implementation details of official Producers (all them have some
behaviour-based clients we can wrap, but not exactly public API). The good thing is this is all optional - just required if span linking is desired.

[1]: https://github.com/open-telemetry/oteps/pull/192
[2]: https://hexdocs.pm/broadway/1.0.3/custom-producers.html


andrewhr:
For the record: while a lot of this is experimental in the sense attributes, methods of propagation etc are WIP, a similar impl is in prod.

tielur:
Any movement on this? Is this blocked by anything? Would love to get open telemetry tracing with broadway

tsloughter:
@andrewhr any update?

andrewhr:
@tsloughter @tielur  sorry for the slow answer. I plan to revisit this implementation this week, as I do need in my current job - mostly keep up to date and give a quick test on a real app.

The caveat @tsloughter [is the name has been captured](https://hex.pm/packages/opentelemetry_broadway) in the meantime. @tomtaylor has been active here, so maybe he can transfer and we could overwrite with a new implementation? I would say what we have in this PR is more complete and deals with span link / propagation so it's worth :)

tsloughter:
Great.

And that'd be great if @tomtaylor agrees?

tomtaylor:
Hi folks - very happy to give this library name over if there's a more complete implementation. I've not properly looked at what @andrewhr has implemented, but on the surface it looks more complete, so let's do it. What's the best way to do that?

tielur:
Looks like we had some good movement on this about 2 weeks ago, I'd love to see this make it to the finish line. Does anyone have steps to take to get this working. 

Do we need this PR merge prior to the library name transfer?

tsloughter:
@tielur name transfer can wait until after merge.

Just waiting on @andrewhr to pull in updates from `main` and mark this as ready instead of `draft`.

yordis:
Closing this as superseded. `main` now includes `opentelemetry_broadway` with later propagation, semantic convention, SQS/PubSub, Broadway 1.3, and batch tracing improvements.

PR #85, opened 2022-05-14
Add Baggage Span Processor

gugahoa:
The purpose of the Baggage Span Processor is to add baggage as attributes to the span

PR #94, opened 2022-06-25
opentelemetry_ecto  - add possibility to setup additional attributes.

fcevado:
current implementation of opentelemetry_ecto doesn't allow to setup additional attributes to the span. this PR add this possibility by allowing an `:attributes` option on setup.

### DIFF
--- a/instrumentation/opentelemetry_ecto/lib/opentelemetry_ecto.ex
+++ b/instrumentation/opentelemetry_ecto/lib/opentelemetry_ecto.ex
@@ -36,7 +36,7 @@ defmodule OpentelemetryEcto do
       defaults to the concatenation of the event name with periods, e.g.
       `"blog.repo.query"`. This will always be followed with a colon and the
       source (the table name for SQL adapters).
-    * `:attributes` - additional attributes to include in the span. If there
+    * `:additional_attributes` - additional attributes to include in the span. If there
       are conflits with default provided attributes, the ones provided with
       this config will have precedence.
   """
@@ -77,7 +77,7 @@ defmodule OpentelemetryEcto do
       end <> if source != nil, do: ":#{source}", else: ""
 
     time_unit = Keyword.get(config, :time_unit, :microsecond)
-    config_attributes = Keyword.get(config, :attributes, %{})
+    additional_attributes = Keyword.get(config, :additional_attributes, %{})
 
     db_type =
       case type do
@@ -106,7 +106,7 @@ defmodule OpentelemetryEcto do
           acc
       end)
       |> Map.merge(base_attributes)
-      |> Map.merge(config_attributes)
+      |> Map.merge(additional_attributes)
 
     parent_context = OpentelemetryProcessPropagator.fetch_parent_ctx(1, :"$callers")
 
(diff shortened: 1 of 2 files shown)


instrumentation/opentelemetry_ecto/lib/opentelemetry_ecto.ex
```diff
@@ -36,6 +36,9 @@ defmodule OpentelemetryEcto do
       defaults to the concatenation of the event name with periods, e.g.
       `"blog.repo.query"`. This will always be followed with a colon and the
       source (the table name for SQL adapters).
+    * `:attributes` - additional attributes to include in the span. If there
```
bryannaegele, line 39: Let's call this additional_attributes. I don't want to overload that word as an option when it comes to optional attributes in the future

instrumentation/opentelemetry_ecto/test/opentelemetry_ecto_test.exs
```diff
@@ -111,7 +124,7 @@ defmodule OpentelemetryEctoTest do
     attach_handler()
 
     try do
-      Repo.all(from u in "users", select: u.non_existant_field)
+      Repo.all(from(u in "users", select: u.non_existant_field))
```
bryannaegele, line 127: It seems like your formatter configuration isn't honoring the formatter file. Please fix that up as there are a lot of formatting changes here unrelated to your change. Thanks!

andrewhr:
Isn't this task better suited for https://github.com/open-telemetry/opentelemetry-erlang-contrib/pull/85?

The concern is this become a pattern for all instrumentation libraries. Not sure how it's the vision for that, specially after the Instrumentation Scope changes. Thoughts @tsloughter?

tsloughter:
@andrewhr that would assume they want the attributes to also be in baggage.

fcevado:
@andrewhr as I understand baggage would be generic for the whole app not particular to a library context. for example, if I wan't to identify write and read replicas, I can specify different `telemetry_prefix` for the repos and setup the instrumentation with specific attributes to them. the same for applications that connect to multiple databases and i want something more descriptive than the `db.url`.

fcevado:
@tsloughter @bryannaegele any update on this?

tsloughter:
I think this is needed but I don't want to be the decided on a change like this to the Ecto integration.

But I'll give an argument for it in principle: in other cases, like an HTTP server span, you can add attributes to the Span the instrumentation creates because the user is writing the code that is being wrapped in a span (so simply call the `set_attribute` macro on the current context). But in the case of an instrumentation like Ecto the control is never given to the user because it simply runs the query.

bryannaegele:
Sorry. I was on vacation when you were still working through those conversations.

I think this is fine at this point? The functionality is required by the spec now. Did you have any specific objection to the implementation @tsloughter?

bryannaegele:
Also, @fcevado you still need to sign the CLA.

tsloughter:
Looks ok to me.

fcevado:
@bryannaegele it's signed :+1:

bryannaegele (approved):

PR #96, opened 2022-07-22
Made 'db.url' more complete and some minor refactor

spencerdcarlson:
Normalize Ecto URI and Config the same way that ecto does it [here](https://github.com/elixir-ecto/ecto/blob/v3.8.4/lib/ecto/repo/supervisor.ex#L28)

Normalizing the URI allows for:
* `db.url` to be more complete
* `net.peer.name` to be added conditionally
* `net.peer.port` to be added conditionally

Using `Ecto.Repo.Supervisor.parse_url/1` to normalize the ecto URI requires that ecto is available at runtime so I added the oldest version of ecto that has what we need ([v0.15.0](https://github.com/elixir-ecto/ecto/blob/v0.15.0/lib/ecto/repo/supervisor.ex#L72-L102)). This seems like a reasonable dependency to add given that these are ecto metrics that are being collected.


instrumentation/opentelemetry_ecto/lib/opentelemetry_ecto.ex
```diff
@@ -125,14 +121,36 @@ defmodule OpentelemetryEcto do
...
+  defp ctx?(_), do: true
+
+  defp maybe_add(map, _, nil), do: map
+  defp maybe_add(map, key, value) when is_map(map), do: Map.put(map, key, value)
+
+  defp maybe_append_source(name, nil), do: name
+  defp maybe_append_source(name, source), do: "#{name}:#{source}"
+
+  defp maybe_add_peer_name(map, peer_name) do
+    if ip?(peer_name), do: map, else: maybe_add(map, "net.peer.name", peer_name)
```
spencerdcarlson, line 143: I'm not sure that this logic is correct. Looking at [opentelemetry-specification](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/span-general.md#netpeername)
I see these two statements:

> Sometimes host name is only available to instrumentation as a string which may contain DNS name or IP address. net.peer.name SHOULD be set to the available known hostname (e.g., "127.0.0.1" if connecting to an URL https://127.0.0.1.com/foo).

> If only IP address is available via net.sock.peer.addr, net.peer.name SHOULD NOT be set.

I think I have this wrong. I think this means we SHOULD always set it in our case regardless of if it is only an ip address.

instrumentation/opentelemetry_ecto/lib/opentelemetry_ecto.ex
```diff
@@ -55,58 +55,54 @@ defmodule OpentelemetryEcto do
...
-    database = repo.config()[:database]
 
-    url =
-      case repo.config()[:url] do
-        nil ->
-          # TODO: add port
-          URI.to_string(%URI{scheme: "ecto", host: repo.config()[:hostname]})
+    %{
+      hostname: hostname,
+      username: username,
```
bryannaegele, line 61: Credentials are not allowed to be exposed in otel data

instrumentation/opentelemetry_ecto/lib/opentelemetry_ecto.ex
```diff
@@ -55,58 +55,54 @@ defmodule OpentelemetryEcto do
...
-    }
+    base_attributes =
+      %{
+        "db.type": if(type == :ecto_sql_query, do: :sql, else: type),
+        "db.statement": query,
+        "db.instance": database,
+        "db.url": url,
+        "total_time_#{time_unit}s": System.convert_time_unit(total_time, :native, time_unit)
+      }
+      |> maybe_add(:source, source)
```
bryannaegele, line 93: Why is this considered optional at this point? We want this attribute to always be present for consistency even if it is an empty string.

instrumentation/opentelemetry_ecto/mix.exs
```diff
@@ -6,7 +6,7 @@ defmodule OpentelemetryEcto.MixProject do
       app: :opentelemetry_ecto,
       description: description(),
       version: "1.0.0",
-      elixir: "~> 1.10",
+      elixir: "~> 1.12",
```
bryannaegele, line 9: We want to support the oldest supported Elixir version we can, so please revert to 1.10.

instrumentation/opentelemetry_ecto/lib/opentelemetry_ecto.ex
```diff
@@ -55,58 +55,54 @@ defmodule OpentelemetryEcto do
...
-      "db.instance": database,
-      "db.url": url,
-      "total_time_#{time_unit}s": System.convert_time_unit(total_time, :native, time_unit)
-    }
+    base_attributes =
+      %{
+        "db.type": if(type == :ecto_sql_query, do: :sql, else: type),
+        "db.statement": query,
+        "db.instance": database,
+        "db.url": url,
```
bryannaegele, line 90: It looks like this is no longer a valid attribute. It's been removed at some point from the spec.

bryannaegele (commented):
Hey, @spencerdcarlson. There's a lot going on in this PR so I addressed several generic things and then we can see if there are any suggestions here that make sense to open in other smaller PRs.

A general note when working on these libraries is we're looking for as much performance as we can reasonably get away with, so it's a best practice to avoid unnecessary enumerations and intermediate data structures wherever we can.

instrumentation/opentelemetry_ecto/lib/opentelemetry_ecto.ex
spencerdcarlson: Looks like it is now `db.connection_string` do we want to set that? [opentelemetry-specification](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/database.md)

instrumentation/opentelemetry_ecto/lib/opentelemetry_ecto.ex
spencerdcarlson: Currently It will either be `nil` or some string value. I was converting `nil` into an "optional" attribute. Do we want to always include it? Should we convert `nil` to `""`?

instrumentation/opentelemetry_ecto/lib/opentelemetry_ecto.ex
```diff
@@ -55,58 +55,54 @@ defmodule OpentelemetryEcto do
...
-      |> Enum.reduce(%{}, fn
-        {k, v}, acc when not is_nil(v) and k in [:decode_time, :query_time, :queue_time, :idle_time] ->
-          Map.put(acc, String.to_atom("#{k}_#{time_unit}s"), System.convert_time_unit(v, :native, time_unit))
-
-        _, acc ->
-          acc
-      end)
+      |> Map.take([:decode_time, :query_time, :queue_time, :idle_time])
+      |> Stream.map(fn {k, v} -> {:"#{k}_#{time_unit}s", System.convert_time_unit(v, :native, time_unit)} end)
+      |> Enum.into(%{})
```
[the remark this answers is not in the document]
spencerdcarlson: I'll change `Map.take/2` to use `Stream.take_while/2` so then we get the both of best worlds, more readable and only one enumeration will take place.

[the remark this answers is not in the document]
bryannaegele: Streams still add overhead to run and are intended for large collections, not a small list. Please revert back to using reduce.

instrumentation/opentelemetry_ecto/lib/opentelemetry_ecto.ex
bryannaegele: Sure. Looks like that's the right thing now.

instrumentation/opentelemetry_ecto/lib/opentelemetry_ecto.ex
```diff
@@ -51,62 +51,62 @@ defmodule OpentelemetryEcto do
...
+      |> Keyword.take([:hostname, :username, :database, :port])
+      |> Enum.into(%{})
 
-        url ->
-          url
-      end
+    url =
+      "ecto://#{hostname}:#{port}/#{database}"
+      |> URI.parse()
+      |> URI.to_string()
```
bryannaegele, line 73: What's the reason to parse the string and then output to a string again?

instrumentation/opentelemetry_ecto/lib/opentelemetry_ecto.ex
```diff
@@ -125,14 +125,43 @@ defmodule OpentelemetryEcto do
...
+  defp maybe_append_source(name, nil), do: name
+  defp maybe_append_source(name, source), do: "#{name}:#{source}"
+
+  defp maybe_add_peer_addr(map, peer_name) do
+    if ip?(peer_name), do: maybe_add(map, :"net.sock.peer.addr", peer_name), else: map
+  end
+
+  defp db_system(Ecto.Adapters.Postgres), do: "postgresql"
+  defp db_system(Ecto.Adapters.MyXQL), do: "mysql"
+  defp db_system(Ecto.Adapters.Tds), do: "mssql"
```
bryannaegele, line 152: These values can be atoms

instrumentation/opentelemetry_ecto/lib/opentelemetry_ecto.ex
```diff
@@ -125,14 +125,43 @@ defmodule OpentelemetryEcto do
...
+  defp db_system(other) when is_atom(other), do: other |> Atom.to_string() |> String.downcase()
```
bryannaegele, line 153: This is not very performant but it would trim the module to the last bit.

```suggestion
  defp db_system(other) when is_atom(other), do: other |> Module.split() |> List.last() |> String.downcase()
```

Ideally, PRs will just be created for any other databases since the list is well-defined in the spec and is only supposed to fall back to a custom value when it isn't in the list.

instrumentation/opentelemetry_ecto/lib/opentelemetry_ecto.ex
```diff
@@ -125,14 +125,43 @@ defmodule OpentelemetryEcto do
...
+  defp db_system(other) when is_binary(other), do: other |> String.downcase()
+  defp db_system(_), do: "unknown"
```
bryannaegele, line 155: This isn't a valid value from the list of [allowed values](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/database.md#connection-level-attributes). Default is probably `other_sql`

tsloughter:
This is looking good now, just 2 suggestions from @bryannaegele and some conflicts to resolve @spencerdcarlson

instrumentation/opentelemetry_ecto/lib/opentelemetry_ecto.ex
spencerdcarlson: To normalize the output

```elixir
hostname = "localhost"
port = 5432
atabase = "db"

"ecto://#{hostname}:#{port}/#{database}" |> URI.parse() |> URI.to_string()
 # => "ecto://myhost:5432/db"

# What if port is "" or nil
port = ""
"ecto://#{hostname}:#{port}/#{database}" |> URI.parse() |> URI.to_string()
# => "ecto://localhost/db"
```

spencerdcarlson:
@tsloughter I believe I resolved all conflicts.

instrumentation/opentelemetry_ecto/lib/opentelemetry_ecto.ex
tsloughter: You don't need to start as a string then. Create a URI from the pieces with schema `ecto` and call `to_string`.

spencerdcarlson:
@tsloughter any update on this?

bryannaegele:
@spencerdcarlson there are still a few outstanding requests that have not been addressed, particularly around adding ecto as a dep and the extra processing around URI strings. Once those have been addressed, I can do a final review with final suggestions/changes.

yordis:
Closing this out. The connection-info goal here (conditionally reporting host/port, matching what Ecto actually connects to) has since landed on `main`, but through a different mechanism than this PR used: instead of adding `ecto` as a runtime dependency and re-parsing the repo's `:url` config with `Ecto.Repo.Supervisor.parse_url/1`, the instrumentation now captures the resolved connection options directly from the `[:ecto, :repo, :init]` telemetry event, then sets `server.address` and conditionally sets `server.port` (skipping known default ports).

`db.url` itself is no longer part of the current DB semantic conventions, so that specific attribute isn't being reintroduced, the equivalent information is now split across `server.address`, `server.port`, and `db.namespace`.

Given the current implementation already covers the underlying need without the extra dependency, and this PR has been stale with unresolved conflicts for a while, I'm going to close it. Thanks for the work here, @spencerdcarlson, it helped inform where the instrumentation ended up.

PR #98, opened 2022-08-10
Fix span context propagation for queries that already have context

jmbutter:
Currently running any Ecto query will set the OpenTelemetry context to
the parent process's context.  That works great for things like
nested preloads where the task is hidden from the user.  But for
intentionally created tasks where the user providers a span context,
we should use that context rather than jumping back to the parent process's
context.

Additionally, after a query runs it was setting the context back to the parent
process's context again, because `OpenTelemetry.Ctx.detach` was called with
the parent process context rather than the token returned from `OpenTelemetry.Ctx.attach`.

This PR fixes both of these issues and includes tests that cover these cases.

### DIFF
--- a/instrumentation/opentelemetry_ecto/lib/opentelemetry_ecto.ex
+++ b/instrumentation/opentelemetry_ecto/lib/opentelemetry_ecto.ex
@@ -102,11 +102,14 @@ defmodule OpentelemetryEcto do
           acc
       end)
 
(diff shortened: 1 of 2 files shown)


bryannaegele:
@jmbutter this was already being addressed in #82. If you have any comments to add or see something missing, please add them and we can get that merged.

jmbutter:
@bryannaegele Thanks, I just saw that too and you beat me to closing this one.  Will do!

PR #103, opened 2022-08-23
Fix bug in OpentelemetryOban.insert_all/4

wizardfrag:
### What does this PR do?
Fixes a bug in OpentelemetryOban.insert_all/4 that causes the request to fail due to OpenTelemetryOban not existing in Oban


tsloughter:
I'm ignorant of how this works, what do you mean by OpenTelemetryOban not existing in Oban and how does it relate to this PR?

wizardfrag:
Sorry @tsloughter, this is an example error:

```
16:20:13.912 [error] ** (RuntimeError) no config registered for OpentelemetryOban instance, is the supervisor running?
(oban 2.13.1) lib/oban/registry.ex:37: Oban.Registry.config/1
(oban 2.13.1) lib/oban.ex:420: Oban.insert_all/5
```

The other functions in this module use the `Oban` instance.

indrekj (approved):
:+1: looks good

Though, from README:

> Oban also supports inserting jobs using `Oban.insert/4`, `Oban.insert_all/2`
and `Oban.insert_all/4`. These are currently not supported by OpentelemetryOban
and are just proxied through to Oban.

So even though it fixes that API, it won't actually add any tracing.

wizardfrag:
> 👍 looks good
> 
> Though, from README:
> 
> > Oban also supports inserting jobs using `Oban.insert/4`, `Oban.insert_all/2`
> > and `Oban.insert_all/4`. These are currently not supported by OpentelemetryOban
> > and are just proxied through to Oban.
> 
> So even though it fixes that API, it won't actually add any tracing.

Ah yep that's very true. I think we use `OpentelemetryOban.insert_all/4` for parity with our other usages of it.

tsloughter (approved):

PR #107, opened 2022-09-14
Optionally disable trace propagation for Tesla

andrewhr:
While we always want spans being produced, trace progation is not desirable in all cases - namely, when calling external parties, as that may leak sensitive information, like one present on Baggage.

This patch introduces a new option `:propagate`, that defaults to `true`.

Some tweaks are made to existing propagation test, fixing how options are used. The approach here is closer to what we see in some middleware tests of Tesla itself.

### DIFF
--- a/instrumentation/opentelemetry_tesla/lib/middleware/opentelemetry_tesla_middleware.ex
+++ b/instrumentation/opentelemetry_tesla/lib/middleware/opentelemetry_tesla_middleware.ex
@@ -2,17 +2,20 @@ defmodule Tesla.Middleware.OpenTelemetry do
   require OpenTelemetry.Tracer
   @behaviour Tesla.Middleware
 
-  def call(env, next, _options) do
+  def call(env, next, opts) do
     span_name = get_span_name(env)
 
     OpenTelemetry.Tracer.with_span span_name, %{kind: :client} do
       env
-      |> Tesla.put_headers(:otel_propagator_text_map.inject([]))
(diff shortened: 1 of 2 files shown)


bryannaegele:
Would you not just have different clients per destination or not include the middleware if not using macros?

tsloughter:
Those options wouldn't resolve the fact that you should be able to use any propagator,not just he global default. Not sure that is defined in the spec yet or not, but its what should be available.

andrewhr:
> Instead I think it should be a propagator option that can override the default global propagator which can then be set to none or undefined to disable propagation.

I didn't considered that approach because... I didn't know was a thing! How that would look like (in pseudo-code)? Those overrides will work per Tesla client? I want *some* requests to propagate and other requests to not propagate, but both should produce traces/metrics etc.

> Would you not just have different clients per destination or not include the middleware if not using macros?

Not sure if I follow. The idea is, I want spans, I just not want to propagate. I need the middleware for that (on current version)

tsloughter:
@andrewhr it would work the same way but set `propagator: none` as the option instead of `propagate: true`.

tsloughter:
Yes, I think override in opts is the way to go. I don't think there should be any needed SDK work for this.

tsloughter:
@andrewhr do you think you'll get to this soon or should I try to take a crack at it?

tsloughter:
I think this is closed by https://github.com/open-telemetry/opentelemetry-erlang-contrib/pull/146/files

PR #109, opened 2022-09-15
Add OpentelemetryMonitor to contrib

derekkraan:
Hi all,

We have been using this in prod for a good number of months now. This has helped us catch a number of issues, since if a process dies before its span can be closed, the span is simply lost.

I thought about how best to get this into opentelemetry. Not 100% sure this is the correct place, but I think this might be a good way to start the discussion, and it's a lower barrier to getting something into the contrib library than into the core library itself.

If there is the feeling that this doesn't belong here then I will release it as a standalone library.

Looking forward to any feedback on this PR.

### DIFF
--- a/instrumentation/opentelemetry_monitor/.formatter.exs
+++ b/instrumentation/opentelemetry_monitor/.formatter.exs
@@ -1,0 +1,4 @@
+# Used by "mix format"
+[
+  inputs: ["{mix,.formatter}.exs", "{config,lib,test}/**/*.{ex,exs}"]
+]
--- a/instrumentation/opentelemetry_monitor/.gitignore
+++ b/instrumentation/opentelemetry_monitor/.gitignore
@@ -1,0 +1,26 @@
+# The directory Mix will write compiled artifacts to.
+/_build/
+
+# If you run "mix test --cover", coverage assets end up here.
+/cover/
+
+# The directory Mix downloads your dependencies sources to.
+/deps/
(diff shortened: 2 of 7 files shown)


bryannaegele:
Thanks Derek! I think we actually want this in the sdk which means it also needs to be in erlang. Any interest in taking this up? 

https://github.com/open-telemetry/opentelemetry-erlang/pull/84

tsloughter:
Yes, would like this in SDK. Do you have interest/time in converting to Erlang? If not there may be someone else who is and we should ask in Slack.

derekkraan:
I think it should be manageable. I'll close this PR and open a new one in the otel-erlang repo once I have converted to Erlang (hopefully next week sometime).

hkrutzer:
@derekkraan I was integrating this in my codebase and I was wondering if there was a specific reason for doing `:ets.insert(__MODULE__, {self(), span_ctx})` inside `monitor()` instead of `handle_call({:monitor, pid})`?

derekkraan:
@hkrutzer the reason is: it's faster.

hkrutzer:
Sure, but aren't you waiting to the handle_call to complete anyway? So the throughput of the single Genserver is still the bottleneck in this situation. I suppose that throughput could increase by having the ETS call be performed outside of the Genserver 🙂

derekkraan:
Yes, I meant throughput.

PR #115, opened 2022-11-01
finch-instrumentation

maocq:
## New Project

Add instrumentation for [Finch](https://hexdocs.pm/finch/Finch.html), an HTTP client with a focus on performance, built on top of Mint and NimblePool.

### Checklist
**Acknowledgements**

* [x]  I certify that I am authorized to assign ownership of this code to the OpenTelemetry project.
* [x]  I acknowledge that ownership of the [Hex](https://hex.pm) package must be given/transferred to the [OpenTelemetry Hex Org](https://hex.pm/orgs/opentelemetry)

**Project Type**

* [x]  Instrumentation
* [ ]  Exporter
* [ ]  Propagator
* [ ]  Metapackage
* [ ]  Example
* [ ]  Util

**Prior Art**

* [ ]  Is this an existing project? Includes private libraries being open sourced.
  
  * [ ]  Yes
  * [x]  No
* [ ]  Are there any other packages on [Hex](https://hex.pm) providing similar functionality?

**Existing Project**

* [ ]  Moving from an existing public repository? Provide a link in the description.
* [ ]  All authors who have contributed code contained in this PR have signed the [CNCF CLA](https://identity.linuxfoundation.org/projects/cncf)

**All Projects**

* [ ]  Label created (maintainers)
* [x]  [labeler.yml](https://github.com/open-telemetry/opentelemetry-erlang-contrib/blob/main/.github/labeler.yml) updated
* [x]  CI
  
  * [x]  Elixir or Erlang language file is updated
  * [x]  Test strategy should use the matrix strategy with all supported language/OTP combinations
  * [x]  The job is scoped to only run for relevant code changes
* [x]  [Codeowners](https://github.com/open-telemetry/opentelemetry-erlang-contrib/blob/main/CODEOWNERS) adds scoped ownership for project maintainers
* [x]  License file (Apache2 - copyright to OpenTelemetry Authors)
* [x]  All code complies with the [Otel Specification](https://github.com/open-telemetry/opentelemetry-specification/tree/main/specification) and Otel Semantic Conventions - [Trace](https://github.com/open-telemetry/opentelemetry-specification/tree/main/specification/trace/semantic_conventions) [Metrics](https://github.com/open-telemetry/opentelemetry-specification/tree/main/specification/metrics/semantic_conventions)


instrumentation/opentelemetry_finch/lib/opentelemetry_finch.ex
```diff
@@ -0,0 +1,73 @@
...
+      "http.scheme": meta.request.scheme,
+      "http.host": meta.request.host,
+      "http.port": meta.request.port,
+      "http.path": meta.request.path,
+      "http.method": meta.request.method,
+      "http.status": status,
+    }
+
+    s =
+      OpenTelemetry.Tracer.start_span("client http", %{
```
tsloughter, line 58: If you start the span at the end then it won't be able to be propagated as trace context in the headers.

instrumentation/opentelemetry_finch/lib/opentelemetry_finch.ex
```diff
@@ -0,0 +1,73 @@
...
+        _ -> 0
+      end
+
+    attributes = %{
+      "http.scheme": meta.request.scheme,
+      "http.host": meta.request.host,
+      "http.port": meta.request.port,
+      "http.path": meta.request.path,
+      "http.method": meta.request.method,
+      "http.status": status,
```
tsloughter, line 54: This should be `status_code`.

Attributes, the span name and the Status of the span need to follow the semantic conventions listed here https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/http.md

tsloughter:
Thanks! There are a few things to clean up around semantic conventions and the ability to propagate but then I think it'll be mergable.

instrumentation/opentelemetry_finch/lib/opentelemetry_finch.ex
maocq: Thank you very much, I have a question, something like this?

```elixir
     parent_context = OpentelemetryProcessPropagator.fetch_parent_ctx(1, :"$callers")

     if parent_context != :undefined do
       OpenTelemetry.Ctx.attach(parent_context)
     end

     s =
       OpenTelemetry.Tracer.start_span(client http, %{ ...
```

tsloughter: Na, I mean calling `start_span` in Finch's request `start` event instead of in `stop`.

maocq: I understand. Since this instrumentation uses telemetry handlers to create OpenTelemetry spans from Finch events, when you want to propagate the trace context in the headers, before making the request, the headers should be set as follows:

```elixir
headers = :otel_propagator_text_map.inject([])

Finch.build(:get, url, headers)
  |> Finch.request(HttpFinch)
```

Do you think that if I put the above information in the documentation, the instrumentation can be a candidate for merge?

Thanks for your time

tsloughter: Ah, blergh, I don't know what I was thinking.

What should happen is the injected trace context is the finch generated span, not its parent. But I guess that isn't possible because telemetry events are being used.

We might just have to settle for this. @bryannaegele ?

The attributes to match the semantic conventions just need to be updated.

instrumentation/opentelemetry_finch/lib/opentelemetry_finch.ex
```diff
@@ -0,0 +1,78 @@
...
+      "http.scheme": meta.request.scheme,
+      "net.peer.name": meta.request.host,
+      "net.peer.port": meta.request.port,
+      "http.target": meta.request.path,
+      "http.method": meta.request.method,
+      "http.status_code": status,
+    }
+
+    s =
+      OpenTelemetry.Tracer.start_span("HTTP #{meta.request.method}", %{
```
tsloughter, line 61: Should also add an option to define a function that accepts the meta struct and returns the span name. But that can wait until after initial merge.

instrumentation/opentelemetry_finch/lib/opentelemetry_finch.ex
tsloughter: Also, yes, including the `inject` in the readme would be good. But it should also note in there that this won't give the trace tree that is expected because the children in the service the request is sent to will be of the parent of the finch request instead of the finch request itself.

This will need to be improved somehow but I guess doesn't need to block this PR. I guess there will need to be a Finch wrapper module, unless Finch incorporated a form of middlewares.

instrumentation/opentelemetry_finch/lib/opentelemetry_finch.ex
maocq: I have already updated the span name and span attributes
Thanks

instrumentation/opentelemetry_finch/lib/opentelemetry_finch.ex
maocq: I just updated the readme. Thank you very much

instrumentation/opentelemetry_finch/README.md
```diff
@@ -0,0 +1,55 @@
...
+*Since this instrumentation is based on telemetry, it is not possible to automatically propagate the context in the http headers. If you need to perform context propagation, you should opt for manual instrumentation. Something like this:*
+
+```elixir
+    require OpenTelemetry.Tracer, as: Tracer
+
+    Tracer.with_span "HTTP #{url}" do
+      headers = :otel_propagator_text_map.inject([])
+
+      case Finch.build(:get, url, headers) |> Finch.request(HttpFinch) do
+        {:ok, %Finch.Response{body: body}} -> {:ok, body}
```
tsloughter, line 38: I suppose it can wait to anohter PR but should really provide a function that'll update the span with the attributes and status based on the Response.

maocq: I agree, I updated the README. Thanks

tsloughter (approved):

maocq:
@tsloughter Thank you very much for the comments and for your time. I only have a couple of questions left

1. Should we do something so that the PR can be merged?
2. We are thinking of sending another PR for the examples folder. It would be an example of instrumentation and metrics in Plug Cowboy and visualization in Zipkin, Jaeger, Prometheus and other tools. Do you think it could be valuable to have such an example?

Thanks

tsloughter:
1. This looks good to merge now.
2. Another example is always welcome!

PR #116, opened 2022-11-09
Add server span kind to otel cowboy

albertored:
This simply set span kind of spans created by the cowboy instrumentation library to SERVER


albertored:
Deps upgraded, now tests should pass

PR #118, opened 2022-11-11
chore: add semantic conventions pkg to tesla

yordis:


instrumentation/opentelemetry_tesla/lib/middleware/opentelemetry_tesla_middleware.ex
```diff
@@ -86,14 +91,14 @@ defmodule Tesla.Middleware.OpenTelemetry do
...
-      "http.host": uri.host,
-      "http.scheme": uri.scheme,
-      "http.status_code": status_code
-    }
+    attrs =
+      %{}
+      |> Map.put(Trace.http_method(), http_method(method))
+      |> Map.put(Trace.http_url(), url)
+      |> Map.put(Trace.http_target(), uri.path)
+      |> Map.put(Trace.net_host_name(), uri.host)
```
yordis, line 99: I am not sure about this one, please double-check me here

instrumentation/opentelemetry_tesla/lib/middleware/opentelemetry_tesla_middleware.ex
```diff
@@ -86,14 +91,14 @@ defmodule Tesla.Middleware.OpenTelemetry do
...
-    attrs = %{
-      "http.method": http_method(method),
-      "http.url": url,
-      "http.target": uri.path,
-      "http.host": uri.host,
-      "http.scheme": uri.scheme,
-      "http.status_code": status_code
-    }
+    attrs =
+      %{}
```
tsloughter, line 95: Should keep this as a single map definition instead of piping through a bunch of `put`s.

instrumentation/opentelemetry_tesla/lib/middleware/opentelemetry_tesla_middleware.ex
tsloughter: Looks right. Looks like there are some additional host attributes in the conventions but those can be added later if they are needed.

instrumentation/opentelemetry_tesla/lib/middleware/opentelemetry_tesla_middleware.ex
bryannaegele: I was going to ask about this. I think the only way to not do that is to use arrow style maps, but that results in a different structure for matching against. Am I missing something?

yordis: I tried. Any idea how it would be?

tsloughter: I don't think it results in anything different:

```
%{Trace.http_method() => http_method(method)
```

Right?

bryannaegele: It's fine. It's just some tests that might need to be adjusted anytime there's a case where users can merge in their own attributes once we implement that in all of these.

yordis: ahhh the rocket syntax ...!!

yordis:
should be ready now 🚀

ricardoccpaiva (approved):

tsloughter:
The formatter test fails saying there is nothing to check, there seems to be some issues with the CI setup.

tsloughter (approved):

PR #119, opened 2022-11-13
chore: add semantic conventions pgk to redix

yordis:


yordis:
I am wondering if `"db.redix.connection_name"` could be replaced or not

tsloughter (approved):

PR #121, opened 2022-11-15
chore: add semantic conventions to phoenix

yordis:


instrumentation/opentelemetry_phoenix/test/opentelemetry_phoenix_test.exs
```diff
@@ -53,16 +53,17 @@ defmodule OpentelemetryPhoenixTest do
...
              "http.status_code": 200,
              "http.target": "/users/123",
-             "http.user_agent": "Mozilla/5.0 (Macintosh; Intel Mac OS X 10.15; rv:81.0) Gecko/20100101 Firefox/81.0",
-             "net.host.ip": "10.211.55.2",
+             "http.user_agent":
+               "Mozilla/5.0 (Macintosh; Intel Mac OS X 10.15; rv:81.0) Gecko/20100101 Firefox/81.0",
+             "net.sock.host.addr": "10.211.55.2",
              "net.host.port": 4000,
-             "net.peer.ip": "10.211.55.2",
+             "net.sock.peer.addr": "10.211.55.2",
```
yordis, line 66: https://github.com/open-telemetry/opentelemetry-specification/blob/120f60ff70ed2a64af26017c8fe611f85e09d13f/CHANGELOG.md?plain=1#LL263C77-L263C95

instrumentation/opentelemetry_phoenix/test/opentelemetry_phoenix_test.exs
```diff
@@ -53,16 +53,16 @@ defmodule OpentelemetryPhoenixTest do
     assert %{
              "http.client_ip": "10.211.55.2",
              "http.flavor": :"1.1",
-             "http.host": "localhost",
+             "net.host.name": "localhost",
```
yordis, line 56: Double-check here, I am not sure if this is correct or not but `http.host` does not exists in the macros

tsloughter: Yea, pretty sure this is right.

instrumentation/opentelemetry_phoenix/lib/opentelemetry_phoenix.ex
```diff
@@ -22,11 +22,13 @@ defmodule OpentelemetryPhoenix do
       end
 
   """
-
-  require OpenTelemetry.Tracer
+  alias OpenTelemetry.SemanticConventions.Trace
```
tsloughter, line 25: I think having it be `SemanticConventions.Trace` would make the attribute keys more clearly convey what they are, like `SemanticConventions.Trace.http_route() => meta.route()`.

`Trace` and `Tracer` are just so similar :). 

But I'm not hardset on it, so would be fine with merging as is as well.

instrumentation/opentelemetry_phoenix/test/opentelemetry_phoenix_test.exs
yordis: Check here, please!

tsloughter: looks right

tsloughter (approved):

PR #123, opened 2022-11-15
chore: ran formatter

yordis:


bryannaegele:
Hey. I'm not sure which directory you ran this from but this doesn't look like it honored any of the formatter files. It's generally run per app and not from the top level

yordis:
@bryannaegele was at the top indeed, I can fix the formatter per project which it requires adding `.formatter.exs` file

yordis:
The changes will make sure that if you run `mix format` form the root, it will respect the nested projects

bryannaegele (approved):

PR #131, opened 2022-11-22
Adding compose file only with dependencies in example

rafaelbpcrespo:
This PR adds a compose file with only the needed dependencies to see the traces of your Elixir project and some documentation around it.


tsloughter:
Thanks! I wonder if we shouldn't make this the default `docker-compose.yml`. That is what I'd prefer.

tsloughter:
@rafaelbpcrespo can you sign the CLA

yordis:
Closing as stale - opened in 2022 with no CLA signed and no activity since. The example's docker-compose.yaml has since evolved into a dependencies-only setup on its own, so this is superseded. Feel free to reopen if you'd like to pick it back up and sign the CLA.

PR #133, opened 2022-11-25
Phoenix Plug.Cowboy adapter support

bryannaegele:
Adding optional adapter support for `Plug.Cowboy` to continue the existing cowboy span rather than starting a new one.

`Bandit.PhoenixAdapter` support will require a telemetry instrumentation lib for it.

The phoenix example app is also updated to be more friendly for user experimentation and iteration by removing the docker app build as the default way to run it.


tsloughter (approved):
I think I get whats going on, so approving :)

instrumentation/opentelemetry_phoenix/lib/opentelemetry_phoenix.ex
```diff
@@ -1,15 +1,41 @@
...
+
+  `Bandit.PhoenixAdapter` is not currently supported.
+
   ## Usage
 
   In your application start:
 
       def start(_type, _args) do
+        :opentelemetry_cowboy.setup()
+        OpentelemetryPhoenix.setup(adapter: :cowboy2)
```
tsloughter, line 38: Is there a way to detect, or just assume, this? Aren't the vast majority of uses cowboy2, why not make it that it is cowboy2 unless you supply an adapter option?

bryannaegele: We'd have to get a config added to specify the endpoint and then parse through the config to find which adapter is in use for it. We can't assume when there are multiple endpoints running because we're setting up the handlers based on the telemetry prefix. The other option is we have to parse through the conn on every request to try to find it which seems a little a 

No matter what, the user is going to have to install server instrumentation for the adapter they're using. I don't have qualms with making the user specify which one to use. It's also likely Bandit will become the goto within a year or two. Its performance is substantially better and has support in Phoenix 1.7. If it holds up, I could see it becoming the default in a future release.

We could make it the default, but if they don't add and setup cowboy instrumentation then it just won't work at all. Felt like a breaking change if we went that route. Or this lib needs to package up multiple adapters and install them whether they use them or not. I mean, Ecto requires you to say which adapter you're using through config. With Ecto repos, you're adding specific repos to the supervision tree to start them. Phoenix does the same with JSON library tied to an endpoint, but the endpoint again is added to the supervision tree. So at some place there's config happening and they're added to the start or supervision tree.

I dunno.

bryannaegele:
Really bad rebase merge so reopening with #144

PR #137, opened 2022-12-04
Add Dataloader instrumentation library

maartenvanvliet:
### What does this PR do?

The [Dataloader](https://hex.pm/packages/dataloader) package is a library often used in conjunction with the Absinthe GraphQL library to perform batch loading of data. This can be useful for efficiently querying databases using Ecto or other custom data sources. It is important to add instrumentation to Dataloader to be able to track and analyze the performance of data loads and identify potential issues such as slow data loads or N+1 queries.

I've looked at the [conventions](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/README.md) for naming the events but could not find any applicable but open to suggestions.


instrumentation/opentelemetry_dataloader/lib/opentelemetry_dataloader.ex
```diff
@@ -0,0 +1,99 @@
...
+      "dataloader.run",
+      metadata,
+      %{
+        kind: :client
+      }
+    )
+  end
+
+  def handle_event(@run_stop, _measurements, metadata, config) do
+    parent_ctx = OpentelemetryProcessPropagator.fetch_parent_ctx(4, :"$callers")
```
tsloughter-splunk, line 59: I found this unexpected. Why is this and the `attach` needed when immediately following the current span is set from metadata and ended?

tsloughter: Right. In your fiddling were you starting a span that is active when Dataloader is called, as in, does the first dataloader span have a parent? I'd expect it not to in the current setup.

olivermt:
Whats the status on this PR? Anything someone can assist with?

instrumentation/opentelemetry_dataloader/lib/opentelemetry_dataloader.ex
```diff
@@ -0,0 +1,99 @@
...
+      "dataloader.run",
+      metadata,
+      %{
+        kind: :client
+      }
+    )
+  end
+
+  def handle_event(@run_stop, _measurements, metadata, config) do
+    OpentelemetryTelemetry.set_current_telemetry_span(config.tracer_id, metadata)
```
tsloughter, line 65: I don't think you need to do this.

tsloughter:
@olivermt it is moving along. I just had another comment, it may be mergable soon.

instrumentation/opentelemetry_dataloader/lib/opentelemetry_dataloader.ex
maartenvanvliet: I got that from here: https://github.com/open-telemetry/opentelemetry-erlang-contrib/blob/fb5598800569cd45d97a891b001f4b39efbaf7d7/instrumentation/opentelemetry_absinthe/lib/opentelemetry_absinthe/execute_handler.ex#L51 but I'l remove it. 

Should it also be removed in the @batch_stop handle_event?

tsloughter: Oh, and now I see its also in Phoenix https://github.com/open-telemetry/opentelemetry-erlang-contrib/blob/main/instrumentation/opentelemetry_phoenix/lib/opentelemetry_phoenix.ex#L136

@bryannaegele you'll have to explain this one to me :)

acrogenesis (approved):

instrumentation/opentelemetry_dataloader/lib/opentelemetry_dataloader.ex
bryannaegele: @tsloughter it's possible another span is open, especially with Phoenix due to when controller functions return, so it was a safeguard to ensure the Phoenix span was closed properly to not lose the trace. I don't know that this is as much of a concern in other libraries, so probably not needed, but also isn't harmful if someone used it.

The other use case would be if some process was creating many spans that are getting closed asynchronously and you therefore have more than one in progress. An example would be a genserver.

I can't speak to the dataloader library, but those are the cases when I'd employ that function.

tsloughter: Ok, yea, I'm guessing I thought `OpentelemetryTelemetry` worked differently when I originally commented.

tsloughter (approved):

PR #65, opened 2022-02-09
Add ecto adapter attribute

jhonndabi:
I couldn't find any discussion about how to include the required attribute `db.system`.
I think currently there isn't a good way to infer that information, maybe ecto adapters can send to telemetry that information through metadata. For now, if we simple pass the ecto adapter as attribute, will be possible to infer and map that info.
`Elixir.Ecto.Adapters.Postgres -> postgresql`
`Elixir.Ecto.Adapters.MyXQL -> mysql`
`Elixir.Ecto.Adapters.Tds -> mssql`

This PR do the following:

- Change some attributes key to follow [semantic conventions](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/database.md).

- Add the attribute `ecto.db.adapter`
- Add the attribute `db.system`

PR #142, opened 2022-12-18
chore: add sc to ecto

yordis:


tsloughter:
Argh. I seemed to have missed this. It now has conflicts and I think will conflict with https://github.com/open-telemetry/opentelemetry-erlang-contrib/pull/65

Trying to get these Ecto PRs all merged but they all have a little bit I'm unsure about.

yordis:
We should try to convince the adapter authors and Ecto to follow some rules to setup such a key.

Although the alternative PR will work, the direction of dependency where OTEL is aware of ALL things out there is brutal.

Maybe the Ecto team could support it a bit and deprecate that way eventually.

tsloughter:
@yordis oh and this is still a draft. Is it still not ready?

yordis:
I mean, kind of yeah. I would probably try to get as many attributes as possible but it seems that #65 has some good stuff there also.

PR #143, opened 2023-01-03
[opentelemetry-tesla] Add an option to not inject_distributed_tracing_headers

enzoqtvf:
Hey team thanks for this wonderful library ! :) 

I added an option inject_distributed_tracing_headers which is set as true by default.

When it's set to false the middleware won't add distributed tracing headers to the request.

This can be useful when you are sending HTTP request to an endpoint you don't own.


tsloughter:
Thanks! What we actually need here is an option to pass the propagator to use (b3, w3c, etc) along with the option `none` which disables injection. Right now it always uses the global default propagator but the user should be able to override that, or disable it with `none`. 

I can't recall the code for that off the top of my head but can help if you need (and you are willing, of course).

enzoqtvf:
@tsloughter Good call, I can dig into that and modify it ! thanks for the suggestion ! I will update the PR in the afternoon :)

tsloughter:
@enzoqtvf any update on this?

tsloughter:
I think this is closed by https://github.com/open-telemetry/opentelemetry-erlang-contrib/pull/143

I'll close it, reopen if necessary.

PR #144, opened 2023-01-05
Phoenix Plug.Cowboy adapter support

bryannaegele:
Adding optional adapter support for `Plug.Cowboy` to continue the existing cowboy span rather than starting a new one.

`Bandit.PhoenixAdapter` support will require a telemetry instrumentation lib for it.

The phoenix example app is also updated to be more friendly for user experimentation and iteration by removing the docker app build as the default way to run it.

Also updates phoenix ecto example which includes running with the latest commits.

Replaces #133


instrumentation/opentelemetry_phoenix/lib/opentelemetry_phoenix.ex
```diff
@@ -131,14 +172,26 @@ defmodule OpentelemetryPhoenix do
   end
 
   @doc false
-  def handle_endpoint_stop(_event, _measurements, %{conn: conn} = meta, _config) do
+  def handle_endpoint_stop(_event, _measurements, meta, _config) do
+    case adapter() do
+      :cowboy2 ->
+        :ok
```
tsloughter, line 178: Why does only the non-cowboy2 version do anything?

bryannaegele: Because otel cowboy will close the span and set all the closing attributes.

tsloughter: Oh. I see now. So this only starts a span if not using cowboy? I had thought a new span was started in the phoenix instrumentation.

tsloughter (approved):

instrumentation/opentelemetry_phoenix/lib/opentelemetry_phoenix.ex
bryannaegele: Correct. It will just continue the existing span which allows it to set the name properly and capture the full request.

PR #146, opened 2023-01-25
Update otel_telemetry.erl

bryannaegele:


utilities/opentelemetry_telemetry/src/otel_telemetry.erl
```diff
@@ -100,7 +100,9 @@ peek_from_tracer_stack(TracerId) ->
         undefined ->
             undefined;
         [SpanCtxSet | _Rest] ->
-            SpanCtxSet
+            SpanCtxSet;
+        [] ->
```
tsloughter, line 104: should there be a debug log?

tsloughter (approved):

PR #154, opened 2023-03-06
[Tesla middleware] `non_error_statuses` option

implicitly-awesome:
This PR adds new `non_error_statuses` option to `Tesla.Middleware.OpenTelemetry`.

Sometimes we expect some HTTP codes not to be considered as errored responses, good example - an API call that returns 404, in some cases receiving 404 is considered as a normal situation. Therefore, we'd like not to have such calls' spans to be marked as "errored".


tsloughter:
Thanks. I think this is probably ok. There was a long discussion about this in the Otel Spec SIG that end with, https://github.com/open-telemetry/opentelemetry-specification/pull/2309#issuecomment-1072476299

So the decision was the user preferably disables these span statuses in the collector.

But being able to do it, while leaving the default as is, in the instrumentation library seems reasonable, in order to not have to setup a collector rule for every single http client requests that you might not want 4xx to be an error.

tsloughter:
Thanks. I think this is probably ok. There was a long discussion about this in the Otel Spec SIG that end with, https://github.com/open-telemetry/opentelemetry-specification/pull/2309#issuecomment-1072476299

So the decision was the user preferably disables these span statuses in the collector.

But being able to do it, while leaving the default as is, in the instrumentation library seems reasonable, in order to not have to setup a collector rule for every single http client requests that you might not want 4xx to be an error.

tsloughter:
After talking to @bryannaegele about this we think a better name and functionality would be if the config is `mark_status_ok` and it marks the status to `ok` instead of leaving unset.

Would marking as `ok` work for your use case?

implicitly-awesome:
hello @tsloughter  👋🏻 

thank you for the feedback, is [that](https://github.com/open-telemetry/opentelemetry-erlang-contrib/pull/154/commits/ab0a5567df7936741a9600d74d59d34892435aee) what you meant? As to me, the prev. `non_error_statuses` option sounded a bit better and was a bit more understandable. WDYT?

tsloughter:
@implicitly-awesome yup!

tsloughter (approved):

PR #156, opened 2023-03-08
[grpcbox] bump version

ty-elastic:
bump release version of grpcbox to include recent, merged changes.

I apologize: how does opentelemetry_grpcbox actually get published to hex (https://hex.pm/packages/opentelemetry_grpcbox)? if we bump the version here, will a github action here (or in opentelemetry-erlang) build and publish it to hex?


tsloughter:
Thanks!

Sadly it is manual for now. I've been meaning to check other repos to see if anyone does a strategy like using a tag prefixed with the package name, `opentelemetry_grpcbox-0.2.0`, and trigger the publish that way, or by checking differences with `main` and if any versions in files changed then publish those...

tsloughter (approved):

PR #160, opened 2023-03-15
OpentelemetryReq plugin

bryannaegele:
This is already published https://hexdocs.pm/opentelemetry_req/OpentelemetryReq.html as I needed it out ASAP for work. 

Everything seems to work well locally but doesn't have any tests. rlopez on Slack said they'd be interested in helping out with that. We can fix up as any errors pop up.


tsloughter:
Is this injecting the parent span context instead of the span context created by this library? That is how I read it, but not positive I got it right.

bryannaegele:
Ugh, yeah. Going to have to do this like ecto :(

tsloughter:
Like Ecto?

tsloughter:
Since Reqs has "steps" can't you start the span in a step before the request is made and end it in the step after the response is received?

bryannaegele:
It requires all the ctx juggling with parent ctx.

tsloughter:
Oh that, so it can't just live in the pdict with `set_current_span`?

bryannaegele:
I don't think end_span from Span or Tracer sets the context back.

bryannaegele:
nm. I got it :)

instrumentation/opentelemetry_req/lib/opentelemetry_req.ex
```diff
@@ -0,0 +1,206 @@
...
+
+    parent_ctx = OpenTelemetry.Ctx.get_current()
+
+    Tracer.start_span(span_name, %{
+      attributes: attrs,
+      kind: :client
+    })
+    |> Tracer.set_current_span()
+
+    Req.Request.put_private(request, :otel_parent_ctx, parent_ctx)
```
tsloughter, line 68: I don't see this used. Did you mean to use it to set the context back to `parent_ctx`?

bryannaegele: Nah. Forgot to delete that. Seems to work fine without juggling.

tsloughter: Eh, it should return to the previous context when it finishes, since it is starting a new one and setting it to the current in the pdict it'd be issue back in the code that called Req.

bryannaegele: So end span would return the previous context when done from `Tracer` and then needs to be attached again?

bryannaegele: The spans currently seem to behave as expected now.

bryannaegele: Ok. Think this is all correct now.

tsloughter (approved):

PR #163, opened 2023-04-15
Improve Req with OTel http spec

rlopzc:
- Span name now follows [http span spec](https://opentelemetry.io/docs/reference/specification/trace/semantic_conventions/http/#name)
  - Span name defaults to `:path_params_template` or if there is no path_params it gets the `request.url.path`
- Http.url now follows [http.url spec](https://opentelemetry.io/docs/reference/specification/trace/semantic_conventions/http/#http-client)
- Override span_name if is provided in the request options

### DIFF
--- a/instrumentation/opentelemetry_req/lib/opentelemetry_req.ex
+++ b/instrumentation/opentelemetry_req/lib/opentelemetry_req.ex
@@ -182,14 +182,6 @@ defmodule OpentelemetryReq do
     end
   end
 
-  defp url(uri) do
-    if uri.query do
-      uri.path <> "?" <> uri.query
-    else
-      uri.path
-    end
-  end
-
   defp http_method(method) do
     case method do
       :get -> :GET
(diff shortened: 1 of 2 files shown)


instrumentation/opentelemetry_req/lib/opentelemetry_req.ex
```diff
@@ -106,12 +122,21 @@ defmodule OpentelemetryReq do
   defp format_exception(_), do: ""
 
   defp span_name(request) do
-    Req.Request.get_private(request, :path_params_template)
+    case request.options[:span_name] do
+      nil ->
+        route = Req.Request.get_private(request, :path_params_template) || request.url.path
```
tsloughter, line 127: This should not use the url path since that is a high cardinality value. The spec says for clients to just use method as the span name.

rlopzc: Oh right, just re-read the spec. Thanks! Addressed

tsloughter (approved):

PR #165, opened 2023-04-19
Record an exception as an error only when Plug.Exception.status/1 returns 500..599.

derekkraan:
It is encouraged in the Phoenix docs to use `Plug.Exception` with custom exceptions to generate for example 404 responses at certain places. These 404s should not be marked as error, since simple "route not found" 404s are also not marked as error.


tsloughter:
I'm good with this, it fits the spec. A little cautious about backwards compatibility, not that we can't change the default, just don't like changing things on people like this.

@bryannaegele are you good with this being merged as is?

bryannaegele:
The spec part is right as far as the error goes but I think we'd still want to record the exception no matter what.

derekkraan:
So just not setting `"error"`? I could live with that.

derekkraan:
Thanks for the feedback. This PR has become very small indeed. Should be good to go now.

tsloughter (approved):

seantanly:
@derekkraan Thanks for the fix!
@tsloughter Any chance this improvement will be released soon? 
The 4xx errors are not playing well with SLO calculations.

Related: https://github.com/open-telemetry/opentelemetry-specification/pull/1998

PR #166, opened 2023-04-22
Don't record DB statements without sanitizaiton

### LINKED ISSUE #155: Ecto: don't record the db statement by default
When this https://github.com/open-telemetry/opentelemetry-specification/pull/3127 is merged the Ecto instrumentation needs to be updated to not record the statement by default.

joshleecreates:
This adds an option to OpentelemetryEcto.setup/1 that allows a query sanitization function to be provided. If it is not provided, queries are not captured (this is the default). This resolves #155 .

### DIFF
--- a/instrumentation/opentelemetry_ecto/test/opentelemetry_ecto_test.exs
+++ b/instrumentation/opentelemetry_ecto/test/opentelemetry_ecto_test.exs
@@ -50,8 +50,6 @@ defmodule OpentelemetryEctoTest do
 
     assert %{
              "db.instance": "opentelemetry_ecto_test",
-             "db.name": "opentelemetry_ecto_test",
-             "db.statement": "SELECT u0.\"id\", u0.\"email\" FROM \"users\" AS u0",
              "db.type": :sql,
              "db.url": "ecto://localhost",
              decode_time_microseconds: _,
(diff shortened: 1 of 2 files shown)


instrumentation/opentelemetry_ecto/test/opentelemetry_ecto_test.exs
```diff
@@ -82,7 +89,6 @@ defmodule OpentelemetryEctoTest do
 
     assert %{
              "db.instance": "opentelemetry_ecto_test",
-             "db.statement": "SELECT p0.\"id\", p0.\"body\", p0.\"user_id\" FROM \"posts\" AS p0",
```
tsloughter, line 85: Can you assert that it isn't there?

bryannaegele:
Maybe I missed it but the option should be something like `db_statement: :disabled (default) | :enabled | fun/1` (pseudo). The sanitization function seems optional.

tsloughter:
How it is now works just by requiring the sanitization function and the user can supply the identify function `fun(X) -> X end` if they want to include the statement without doing any sanitizing -- instead of having it be 3 possible values for the option.

I'm fine either way.

bryannaegele:
Under the hood we could use that as the function but I don't think it
should be incumbent on the user to have to provide it themselves. Doesn't
feel like a good API design. I didn't read all of the linked otel issue but
the gist seemed like providing a function was a third option to provide,
not just always disabled or function.

tsloughter:
Yea, I was thinking it was purposely requiring the user be extra explicit because they are doing something potentially "dangerous".

bryannaegele:
Yeah, I didn't see the danger part. It's a templated query if done properly but could be a large attribute value on a large query statement. We've got a few of those. I suppose some folks could be running some non-templated query. 🤷🏻

tsloughter:
Yea, the worry is non-templated queries.

joshleecreates:
I can see the value in having an `:enabled` option with `:disabled` being the default — happy to update this if that is the consensus.

tsloughter:
Once the formatting issues and conflicts are resolved I'll merge this.

joshleecreates:
Sounds good, I've been OOO but I can clean this up this week.

PR #168, opened 2023-04-25
Add source_url_pattern to enable linking to source

spatrik:
Just adding the `source_url_pattern` option for exdoc to opentelemetry_oban, to enable the "Link to source" button in the docs.


tsloughter:
Happy to merge this, but figured I should ask if you'd be able to do this for the other apps in the repo :)

spatrik:
Sure! It became a small yak shave expedition. I've now also 

* Add README.md as an "extra" where it wasn't already
* Add a `main` setting. They all have a very obvious main module. Set
  that as `main`, so a user is shown this immediately instead of a list
  of usually only this module.

tsloughter (approved):

PR #176, opened 2023-05-24
[opentelemetry_tesla] Conditionally override propagator

nirev:
This builds on the work of #107 and adds the recommended approach that was mentioned there: setting `propagator: :none` instead of `propagate: false`

leaving room for extending it to support multiple propagators


nirev:
@tsloughter I saw a previous comment you made on https://github.com/open-telemetry/opentelemetry-erlang-contrib/pull/143#issuecomment-1369684774 and I don't know how to proceed.

> What we actually need here is an option to pass the propagator to use (b3, w3c, etc) along with the option none which disables injection.

Where are those propagators (b2, w3c, etc) defined?

nirev:
@tsloughter I think I got it now :)

nirev:
this is ready for review!
let me know if it needs any changes

tsloughter (approved):

PR #180, opened 2023-06-16
Redis: propagate the span context if available

tomtaylor:
Behave in a similar way to Ecto, which is to attach the context of the connection, falling back to the calling process context if available.


instrumentation/opentelemetry_redix/lib/opentelemetry_redix.ex
```diff
@@ -63,6 +63,22 @@ defmodule OpentelemetryRedix do
       |> Map.merge(net_attributes(connection))
       |> Map.merge(redix_attributes(meta))
 
+    parent_context =
+      case OpentelemetryProcessPropagator.fetch_ctx(self()) do
```
tsloughter, line 67: Isn't this just the same as checking `OpenTelemetry.Ctx.get_current()`?

tsloughter: I see it returns undefined instead of an empty context.

tsloughter (approved):

instrumentation/opentelemetry_redix/lib/opentelemetry_redix.ex
tomtaylor: Yeah, this is the same behaviour as `opentelemetry_ecto`

Issue #183, opened 2023-07-08
HTTP client/server instrumentations should have an option to include headers as attirbutes


tsloughter:
Each HTTP client/server instrumentation in this repo needs to add an option for including headers as attributes that follows the semconv:

https://github.com/open-telemetry/semantic-conventions/blob/main/docs/http/http-spans.md#http-request-and-response-headers

albertored:
I can start with the cowboy instrumentation

PR #184, opened 2023-07-08
Allow to configure which headers should be used as span attrs

albertored:
See #183 

~~I implemented this on top of #128 because it was more convenient for me so we should wait for merging that PR before merging this one.~~


tsloughter:
Thanks! Pls update the CHANGELOG tho :)

instrumentation/opentelemetry_cowboy/src/opentelemetry_cowboy.erl
```diff
@@ -139,3 +149,46 @@ client_ip(Headers, RemoteIP) ->
...
+
+set_header_attribute(Header, Value, HeadersAttributes) when is_list(Value) ->
+    maps:put(binary_to_atom(<<"http.request.header.", Header/binary>>), Value, HeadersAttributes);
+set_header_attribute(Header, Value, HeadersAttributes) ->
+    set_header_attribute(Header, [Value], HeadersAttributes).
+
+normalize_header_name(Header) when is_binary(Header) ->
+    normalize_header_name(binary_to_list(Header));
+normalize_header_name(Header) when is_list(Header) ->
+    iolist_to_binary(string:replace(string:to_lower(Header), "-", "_", all)).
```
tsloughter, line 194:
```suggestion
    unicode:characters_to_binary(string:replace(string:to_lower(Header), "-", "_", all)).
```

albertored:
Marked as draft since it need #186 to be merged before proceeding

bryannaegele (commented):
This should all be covered now.

PR #185, opened 2023-07-09
[phoenix instrumenter] Allow to configure which headers should be used as span attrs

albertored:
See #183


tsloughter:
Thanks! Pls update the CHANGELOG tho

albertored:
Marked as draft since it need #186 to be merged before proceeding

PR #186, opened 2023-07-14
Opentelemetry instrumentation http

albertored:
After a discussion about #183 we thought it would be nice to add a utility library for HTTP instrumentation.

At the moment I included in the lib only the code needed for extracting headers into attributes but it can be extended in the future to add more common functionalities.

Names of the options that each HTTP instrumentation library should expose (for instance to take a list of headers that should be extracted into span attributes) are not included on the library. We can put them here as macros but in that case, we should extend the lib and make it a mixed elixir/erlang one. I think it is an overkill for now but let me know what you think.

I added also the CI stuff (hope it works),


tsloughter:
:tada:

utilities/opentelemetry_instrumentation_http/src/opentelemetry_instrumentation_http.erl
```diff
@@ -0,0 +1,108 @@
...
+-ifdef(TEST).
+-include_lib("eunit/include/eunit.hrl").
+-endif.
+
+-export([
+    extract_headers_attributes/3,
+    normalize_header_name/1
+]).
+
+-spec normalize_header_name(Header :: binary() | string()) -> binary().
```
tsloughter, line 12:
```suggestion
-spec normalize_header_name(binary()) -> Result when
      Result :: binary() | {error, string(), RestData} | {incomplete, binary(), binary()},
      RestData :: latin1_chardata() | chardata() | external_chardata().
```

tsloughter:
I think this is a good first start and just needs to have its changelog reflect what it has in the first commit.

utilities/opentelemetry_instrumentation_http/docs.sh
```diff
@@ -0,0 +1,15 @@
...
+# Setup:
+#
+#     mix escript.install github elixir-lang/ex_doc
+#     asdf install erlang 24.0.2
+#     asdf local erlang 24.0.2
+
+rebar3 compile
+rebar3 as docs edoc
+version=0.1.0
+ex_doc "opentelemetry_cowboy" $version "_build/default/lib/opentelemetry_instrumentation_http/ebin" \
```
tsloughter, line 13:
```suggestion
ex_doc "opentelemetry_http" $version "_build/default/lib/opentelemetry_instrumentation_http/ebin" \
```

tsloughter (approved):

albertored:
@tsloughter how does it work the procedure for publishing this on hex.pm? Once published I can proceed with using it in instrumentation libraries

PR #187, opened 2023-07-25
Include `db.system` attribute in `opentelemetry_ecto`

tomtaylor:
Added clauses for the major Ecto adapters, and refactored how attributes are assembled, to make it easier to read.


tsloughter (approved):
Looks good to me. @bryannaegele any thoughts?

instrumentation/opentelemetry_ecto/lib/opentelemetry_ecto.ex
```diff
@@ -90,9 +90,19 @@ defmodule OpentelemetryEcto do
         _ -> type
       end
 
+    db_system =
+      case repo.__adapter__() do
+        Ecto.Adapters.Postgres -> :postgresql
+        Ecto.Adapters.MyXQL -> :mysql
+        Ecto.Adapters.SQLite3 -> :sqlite
+        Ecto.Adapters.Tds -> :mssql
+        _ -> :other_sql
```
tsloughter, line 99: This isn't guaranteed, right? A Repo could be any storage?

tomtaylor: Yes, true. Maybe better falling back to `nil` then?

andrewhr: What about default to `:other_ecto`? This should avoid a `nil` on a required attribute.

tomtaylor: Yeah, makes sense to me. And if there’s no value we should just no send the property at all, right? Rather than a `nil`?

andrewhr: The `ecto` (or `other_ecto`) instead of nil, given it's a required field is too much a stretch?  What do you think, @tsloughter?

tomtaylor: Shall we just drop this attribute if it doesn't match right now? For 99% of users that gets this attribute in and working, and we can revisit the fallback in the future?

tsloughter: Yea, probably the best way forward.

andrewhr (approved):

andrewhr:
Can't wait for it to get merged @tomtaylor, great work! 🚀

tsloughter (approved):

tsloughter:
Will probably pass once I merge 188 and then will merge this.

tsloughter:
Will merge after the conflicts are resolved.

tsloughter:
@tomtaylor ping

tomtaylor:
This should be ready to go now.

tsloughter:
Thanks!

PR #188, opened 2023-07-26
Improve test matrix and add support for Elixir 1.15 and OTP 26

tomtaylor:
I wanted to prove one of the libraries worked on OTP 26 and Elixir 1.15, but hit a wall of errors. I had a go at fixing most of them.

- Implemented the matrix being read from the files, as was originally intended (from what I can tell)
- Added support for building on OTP 26 and Elixir 1.15, including upgrading the dependencies on projects which require `ssl_verify_fun` as only >= 1.1.7 builds on Elixir 1.15
- I added support for only running `mix format --check-formatted` on a specific version of Elixir - the formatter implementation changed subtly between 1.14 and 1.15, so tests will fail

Outstanding issues:

- Some libraries require 1.12 or 1.14 minimum, but we still attempt to run the entire matrix on them. We should conditionally check whether the environment is new enough to support the project before running the workflow.


tsloughter:
Hm, wonder if there is a way to make this actually run the tests on all these that it effects. @bryannaegele any idea?

bryannaegele:
> Hm, wonder if there is a way to make this actually run the tests on all these that it effects. @bryannaegele any idea?

You or I have to trigger the build because they aren't an org member. You can just remove/add `scope-ci` to kick it.

tsloughter:
Ah cool, thanks.

So yea, some of these fail, haven't looked at why yet.

tomtaylor:
There’s a range of reasons why the remaining stuff fails, but from what I can see there’s two groups:

- Libraries which set minimum Elixir version requirements but are being run with a test matrix of lower version requirements and immediately bailing out
- Libraries which have actually failing tests or compilation warnings which should be addressed in the library code

For the first group, we could read the minimum version from `mix.exs` and check whether it meets the running version and bail out early if not.

tsloughter:
We should have a minimum version of elixir for all libraries. I think we were planning on the last 2 major releases, so 1.15 and 1.14?

If the test matrix was just 1.15 and 1.14 would that work then? (Aside from the library issues we need to resolve separately)

bryannaegele:
I think we were trying to match the support levels we have in beam-telemetry. Whatever the "official" levels are for Elixir. I think it's in the Elixir docs.

tomtaylor:
@tsloughter I don't think any libraries require > 1.14, so setting the matrix to 1.14 + 1.15 would work for this class of failing test.

tomtaylor:
OK, I think that's everything passing now. I've fixed a bunch of minor issues in libraries, and aligned all the formatting with 1.15. We now only test on 1.14 and 1.15 and a matrix of recent OTP versions.

.github/elixir-test-matrix.json
```diff
@@ -1,29 +1,14 @@
...
-  ],
-  "os": [
-    "ubuntu-20.04"
-  ],
-  "exclude": [
-    {
-      "elixir_version": "1.11.4",
-      "erlang_version": "25.1.2"
-    },
+  "otp_version": ["26.x", "25.x", "24.x"],
```
bryannaegele, line 2: Versions need to be pinned to the furthest extent possible because it messes with the caches becoming broken anytime a new version is released. We don't really care if it's not running the absolute newest patch for everything.

instrumentation/opentelemetry_ecto/mix.exs
```diff
@@ -6,7 +6,7 @@ defmodule OpentelemetryEcto.MixProject do
       app: :opentelemetry_ecto,
       description: description(),
       version: "1.1.1",
-      elixir: "~> 1.10",
+      elixir: "~> 1.14",
```
bryannaegele, line 9: We should always leave this as low as the library will allow. Unless we require a feature from a newer version, we should go as low as possible.

This note applies to all the libraries.

bryannaegele:
@tsloughter I asked Jóse what he thought since I think we want to try to keep support as close to our beam-telemetry support which is currently 5 Elixir versions which means covering OTP 21 which is still the problematic one. Do we want to adopt a formal policy on last 3 OTP versions? That would take us to OTP 23+ and a min Elixir support of 1.11.4. He seemed good with that.

I just want to align this repo as much as possible to what we want to formally support in otel/otel api.

https://hexdocs.pm/elixir/compatibility-and-deprecations.html

tsloughter:
@bryannaegele yea, 3 is good, that is what we do for rebar3 as well.

bryannaegele:
Ok. Let's actually do 4 and match what we support in otel/otel api. I'd like anything in contrib to be usable with the main libraries.

@tomtaylor we will want to update the matrix then to have Elixir versions down to 1.11.4 and OTP 23 and correct anything from there.

tomtaylor:
I've pinned Elixir + OTP to specific versions and included the versions discussed, and I suspect there are still some failing libraries, but I don't have permissions add/remove labels to fix Github Actions to run them. Could someone take a look?

bryannaegele (approved):

tsloughter (approved):

tsloughter:
argh, some setup-beam error as usual :) https://github.com/open-telemetry/opentelemetry-erlang-contrib/actions/runs/5974597901/job/16209053026?pr=188

tomtaylor:
> argh, some setup-beam error as usual :) https://github.com/open-telemetry/opentelemetry-erlang-contrib/actions/runs/5974597901/job/16209053026?pr=188

That looks to me like the server returned an error looking up that version. Every job in the workflow requests rebar 3.20 and installed it correctly, bar that one. I suspect it'd work on a retry.

tsloughter:
Formatting issue https://github.com/open-telemetry/opentelemetry-erlang-contrib/actions/runs/5974597901/job/16210352781?pr=188

tomtaylor:
I think that landed with https://github.com/open-telemetry/opentelemetry-erlang-contrib/pull/198/commits/6e40de665a8ef39626533ceb74e93db0792dde07

tsloughter:
Ah, must have been added when I ran the `update` on this PR.

Can you run the formatter again and update this PR?

tomtaylor:
Done and all passing - quick, merge it in before something else lands!

PR #189, opened 2023-07-27
Add abilty to hide `db.statement` in `opentelemetry_redix`

tomtaylor:
It's not always necessary/desired to include this attribute, especially as in some cases it can be very long.

### DIFF
--- a/instrumentation/opentelemetry_redix/lib/opentelemetry_redix.ex
+++ b/instrumentation/opentelemetry_redix/lib/opentelemetry_redix.ex
@@ -30,8 +30,12 @@ defmodule OpentelemetryRedix do
 
   You may also supply the following options:
 
-    * `db_statement` - a boolean used to set whether a sanitized `db.statement`
-      attribute is included in each span. Defaults to `true`.
+    * `db_statement` - `:enabled`, `:disabled` or a single arity function. If
+      `:enabled`, the DB statement is sanitized of sensitive values, such as
+      AUTH crendentials, but otherwise logged. If `:disabled`, the statement is
+      not logged. If you pass in a single arity function, the list of commands
+      will be passed through that function for your own processing. Defaults to
+      `:enabled`.
   """
   @spec setup(opts()) :: :ok
   def setup(config \\ []) do
@@ -45,7 +49,7 @@ defmodule OpentelemetryRedix do
 
   @doc false
   def handle_pipeline_stop(_event, measurements, meta, config) do
-    add_db_statement = Keyword.get(config, :db_statement, true)
+    add_db_statement = Keyword.get(config, :db_statement, :enabled)
 
     duration = measurements.duration
     end_time = :opentelemetry.timestamp()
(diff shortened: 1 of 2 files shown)


andrewhr:
This is defined [on documentation](https://opentelemetry.io/docs/specs/otel/trace/semantic_conventions/database/#call-level-attributes) as a recommended attribute. Do you have examples of other libraries (java, node, ruby etc) that could optionally remove it?

Also, is this causing issues? For example, Ecto statements tends to be even bigger than (most) Redis, still there is no configuration to opt-out this attribute.

Sorry to push back a bit like that, just want to understand more the use case and keep as consistent as possible with prior art from OTEL community.

Thanks in advance @tomtaylor!

tsloughter:
The docs actually say it should only be collected if the statement is sanitized.

I forget where we are at with Ecto but it should not be collected by default because we do not sanitize.

tsloughter:
Ah right, its in, https://github.com/open-telemetry/opentelemetry-erlang-contrib/pull/166

tomtaylor:
I’m away at the moment without my laptop, but a cursory look showed me that the Ecto instrumentation supports a custom sanitiser as mentioned in #166. The Ruby Redis library supports something similar.

Our example is that we do a `KEYS` request for a _lot_ of very long patterns at a regular interval, and this is logged verbatim.

Happy to turn this PR into a custom sanitiser if that’s what’s commonly implemented across multiple languages and there’s prior art in Elixir. Does that seem like a better approach?

tsloughter:
@tomtaylor yup, that'd be great

andrewhr:
@tomtaylor @tsloughter just a heads up, I'm on finishing points with a mongo instrumentation and following that same pattern, with a :disabled, :plain, :obfuscated (basic sanitization) and custom fun options.

Sounds like a good plan to me!

tsloughter:
@tomtaylor any update on the switch to a function from a boolean?

tomtaylor:
I haven't had time to dig into this yet. Happy for someone else to chip it in, or I'll try and get to it when I've got a bit of time.

tomtaylor:
@tsloughter I've just sorted this and moved from a boolean config option to `:enabled`, `:disabled` or a single arity function, to mimics how the Ecto instrumentation behaves.

instrumentation/opentelemetry_redix/lib/opentelemetry_redix.ex
```diff
@@ -27,19 +27,30 @@ defmodule OpentelemetryRedix do
 
   @doc """
   Initializes and configures the telemetry handlers.
+
+  You may also supply the following options:
+
+    * `db_statement` - `:enabled`, `:disabled` or a single arity function. If
```
whatyouhide, line 33: Nit:

```suggestion
    * `:db_statement` - `:enabled`, `:disabled` or a single arity function. If
```

instrumentation/opentelemetry_redix/lib/opentelemetry_redix.ex
```diff
@@ -27,19 +27,30 @@ defmodule OpentelemetryRedix do
 
   @doc """
   Initializes and configures the telemetry handlers.
+
+  You may also supply the following options:
```
whatyouhide, line 31: My suggestion: I would either go with `:sanitized_and_logged | :not_logged | fun`, or just `true | false | fun`.

instrumentation/opentelemetry_redix/lib/opentelemetry_redix.ex
```diff
@@ -27,19 +27,30 @@ defmodule OpentelemetryRedix do
...
+
+    * `db_statement` - `:enabled`, `:disabled` or a single arity function. If
+      `:enabled`, the DB statement is sanitized of sensitive values, such as
+      AUTH crendentials, but otherwise logged. If `:disabled`, the statement is
+      not logged. If you pass in a single arity function, the list of commands
```
whatyouhide, line 36: My suggestion is to have a type for the function to point to and document better. For example:

```elixir
@type db_statement_fun() :: (Redix.command() -> String.t())
```

yordis:
Superseded by #660, which added `opt_out_attrs: [DBAttributes.db_query_text()]` to fully disable the query-text attribute, on top of the existing per-command sanitization that already redacts literal values by default. Closing.

PR #192, opened 2023-08-18
opentelemetry_req: Don't assume `request.options` is a map

wojtekmach:
I'm planning to make `request.options` internal representation private (https://github.com/wojtekmach/req/pull/222) though calling `request.options[key]` is still explicitly allowed. I believe this PR keeps the current behaviour for regular usage. The minor change is now we would _not_ raise if someone set `path_params: nil` but note sure how common that'd be. If that's a concern I believe `!is_list(request.options[:path_params])` or `Access.fetch(request.options, :path_params) == :error` would work too.


bryannaegele (approved):
Thanks @wojtekmach!

PR #193, opened 2023-08-19
opentelemetry_req: Don't assume `request.headers` shape

wojtekmach:
I'm planning to switch Req [headers to be maps](https://github.com/wojtekmach/req/issues/223). This patch will work on either representation. After we switch to maps, `Req.Request.put_header/3` should be plenty fast, however I'd be happy to send another patch that updates the headers in bulk if that'd be more desirable.

cc @bryannaegele


wojtekmach:
I just noticed there's a couple ongoing PRs:

- #184 (cc @albertored)
- #194 (cc @andrewhr, thanks for pointing out possible conflicts!)

I'm happy to wait until they are addressed and rebase on top. :)

Just FYI, the :headers change would only happen in Req v0.4.0 which is planned to happen in the next couple of weeks. Ideally :opentelemetry_req would have depended on `{:req, "~> 0.3.0"}`, that is, projects using both dependencies would not have picked up req v0.4.0 when it comes out. But I think it's gonna be ok, if we can get a new opentelemetry_req version with a patch like this one, I'm happy to mention the need to upgrade in my Req v0.4 changelog. Sorry for the extra hassle, if there's anything I can do to smooth the transition on req or opentelemtry_req, please let me know!

instrumentation/opentelemetry_req/test/opentelemetry_req_test.exs
```diff
@@ -1,4 +1,64 @@
...
+
+    resp =
+      Req.get!(req,
+        adapter: adapter,
+        url: "/users/:id",
+        path_params: [id: 3]
+      )
+
+    assert resp.status == 204
+    assert_receive {:span, span(name: "GET /users/:id")}
```
bryannaegele, line 41: Updating this to match pre 1.18 otel spec behavior until that's tackled.

```suggestion
    assert_receive {:span, span(name: "/users/:id")}
```

bryannaegele (commented):
@wojtekmach and @andrewhr I'm going to go with this PR with the tests as-is in here so we can get a release out to support >= 0.4

bryannaegele (approved):

PR #194, opened 2023-08-21
Add tests and tweak behavior of opentelemetry_req

andrewhr:
The main objective of this patch is to add few tests to cover main points of how this library handles OpenTelemetry.

As by-product of those tests, few tweaks were done to production code. More specifically, how we handle `:path_params` validation.

`:path_params` only has restrictions when used for span names, due constraints of "span names should never be unbounded". As such, when we set `:span_name` with a custom one, the check is skipped.

Another change is how `:no_path_params` imply: here, I understand that we as library users are asserting that "I comply and abide this path has no dynamic parts", and as such, it's safe to just use it instead of fallback to only METHOD. Otherwise, we would need some workaround like set empty :path_params just to ensure "safe" paths are reported correctly on span names.

I've considered addting Yet Another Option for "I don't want any path inference" that disables check and set span names just as METHOD, but refrain to do so here to not deviate too much from current implementation.


andrewhr:
Just realized when prep this PR for publication that @wojtekmach has some extra patches, in case of any conflicts.

instrumentation/opentelemetry_req/test/opentelemetry_req_test.exs
```diff
@@ -1,4 +1,120 @@
...
+  end
+
+  for {name, spec} <- Record.extract_all(from_lib: "opentelemetry_api/include/opentelemetry.hrl") do
+    Record.defrecord(name, spec)
+  end
+
+  setup do
+    :otel_simple_processor.set_exporter(:otel_exporter_pid, self())
+
+    :ok
```
wojtekmach, line 19: you may also consider returning:

```elixir
    pid = self()

    adapter = fn req ->
      send(pid, {:req, req})
      {req, Req.Response.new(status: 204)}
    end

    req =
      Req.new(adapter: adapter)
      |> OpentelemetryReq.attach()

    %{req: req}
```

which would remove some duplication from the tests. But then again such duplication doesn't hurt too much either. :)

wojtekmach (commented):
Thanks for the heads up! Your PR is more comprehensive so I'm happy to wait until it's addressed and rebase mine on top. :)

instrumentation/opentelemetry_req/lib/opentelemetry_req.ex
```diff
@@ -115,19 +114,16 @@ defmodule OpentelemetryReq do
...
+  defp format_error(error), do: inspect(error)
 
   defp span_name(request) do
     case request.options[:span_name] do
       nil ->
         method = http_method(request.method)
 
         case Req.Request.get_private(request, :path_params_template) do
-          nil -> method
+          nil -> "#{method} #{request.url.path}"
```
bryannaegele, line 126: We aren't doing 1.20+ spec piecemeal, so please keep this as it was. All the libs will have to be updated at one time.

The template on line 127 was prematurely merged and not reverted yet, so I can't publish until both are reverted to the behavior that was present prior to that PR.

instrumentation/opentelemetry_req/test/opentelemetry_req_test.exs
```diff
@@ -1,4 +1,120 @@
...
+      Req.new(adapter: fn req -> {req, Req.Response.new()} end)
+      |> OpentelemetryReq.attach()
+      |> Req.get(url: "http://example.com/users", span_name: "GET")
+
+    assert_receive {:span, span(name: "GET", kind: :client, attributes: attributes)}
+
+    assert %{
+             "http.url": "http://example.com/users",
+             "http.method": :GET,
+             "http.scheme": "http",
```
bryannaegele, line 33: This should be an atom.

bryannaegele:
> "I comply and abide this path has no dynamic parts", and as such, it's safe to just use it instead of fallback to only METHOD.
The fallback is just whatever the path is in the request, not method. Maybe that changes with 1.20+ spec but until then the behavior is just to use the path, e.g. `/api/users`

instrumentation/opentelemetry_req/lib/opentelemetry_req.ex
bryannaegele: The line in question is from https://github.com/open-telemetry/opentelemetry-erlang-contrib/commit/19a44fbd6830ae71ab161028d67ce6ae40091766

tsloughter: I've forgotten the plan for 1.20, hehe. Did we write it down somewhere. I vaguely remember. Maybe we should do it "now" if there are ppl looking to contribute towards the goal (not saying this should be in this PR but maybe @andrewhr can open a separate one).

bryannaegele: I'd like to do it all at once with 1.20 semconv. What do you think?

bryannaegele: #197

andrewhr: From my understanding of older versions, the "best effort" of span names were always there, so this doesn't not necessary mark a breaking change from the convention definitions. In other words, was not my intend here to piecemeal implement newer specs. Having said that, I totally understand if we prefer to rollback this change - it just a matter of ergonomics so Spans can be consistent in cases we can assert they aren't dynamic.

Btw, thanks for the write up on #197, that's quite helpful! 🙇

instrumentation/opentelemetry_req/test/opentelemetry_req_test.exs
andrewhr: That's just what `URI.parse/1` returns, and which is the same behavior of Tesla middleware. Should we still change it?

I thought I have is, in case we still prefer atoms (perf reasons, right?), I can map to few known schemes and fallback to return as string. Does that seems a good compromise?

bryannaegele: It can just be to_atom but we can do that in another PR if it's a big lift. The [list of registered schemes](https://www.iana.org/assignments/uri-schemes/uri-schemes.xhtml) is massive.

bryannaegele:
#375 should cover everything

PR #195, opened 2023-08-21
Improve Oban plugin handler and fix error handling

andrewhr:
The main purpose of this patch is to introduce plugin aware attributes - mostly counts of processed jobs.

Plugins cover all the offerings of Oban 2.15 and Oban Pro 1.0, the current versions at time of writing.

While testing this in a real app, I've noticed issues with exception handlers for plugin. Decided to slip in this patch changes to both job and plugin event handlers to rely on standard :telemetry.span/3 attributes instead of :error key (which seems like an artifact of earlier versions).

### DIFF
--- a/instrumentation/opentelemetry_oban/lib/opentelemetry_oban/plugin_handler.ex
+++ b/instrumentation/opentelemetry_oban/lib/opentelemetry_oban/plugin_handler.ex
@@ -3,64 +3,127 @@ defmodule OpentelemetryOban.PluginHandler do
 
   @tracer_id __MODULE__
 
-  def attach() do
-    attach_plugin_start_handler()
-    attach_plugin_stop_handler()
-    attach_plugin_exception_handler()
-  end
+  @plugin_start [:oban, :plugin, :start]
+  @plugin_stop [:oban, :plugin, :stop]
+  @plugin_exception [:oban, :plugin, :exception]
 
-  defp attach_plugin_start_handler() do
-    :telemetry.attach(
-      "#{__MODULE__}.plugin_start",
-      [:oban, :plugin, :start],
-      &__MODULE__.handle_plugin_start/4,
-      []
+  def attach do
+    :telemetry.attach_many(
+      [__MODULE__, :plugin],
+      [
+        @plugin_start,
+        @plugin_stop,
+        @plugin_exception
+      ],
+      &__MODULE__.handle_event/4,
+      _config = %{}
     )
   end
 
-  defp attach_plugin_stop_handler() do
-    :telemetry.attach(
-      "#{__MODULE__}.plugin_stop",
-      [:oban, :plugin, :stop],
-      &__MODULE__.handle_plugin_stop/4,
-      []
-    )
+  @doc false
+  def handle_event(event, measurements, metadata, config)
+
+  def handle_event(@plugin_start, _measurements, %{plugin: plugin} = metadata, _config) do
+    span_name = "#{inspect(plugin)} process"
+
+    attributes = %{
+      "messaging.oban.plugin": inspect(plugin)
+    }
+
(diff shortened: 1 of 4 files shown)


tsloughter:
I don't know anything about Oban plugins. @indrekj can you comment on this? @bryannaegele any thoughts?

instrumentation/opentelemetry_oban/lib/opentelemetry_oban/plugin_handler.ex
```diff
@@ -3,64 +3,127 @@ defmodule OpentelemetryOban.PluginHandler do
...
-  defp attach_plugin_exception_handler() do
-    :telemetry.attach(
-      "#{__MODULE__}.plugin_exception",
-      [:oban, :plugin, :exception],
-      &__MODULE__.handle_plugin_exception/4,
-      []
-    )
-  end
+  def handle_event(@plugin_start, _measurements, %{plugin: plugin} = metadata, _config) do
+    span_name = "#{inspect(plugin)} process"
```
indrekj, line 27: Is the inspect intentional? It will add quotes around the plugin name.

instrumentation/opentelemetry_oban/lib/opentelemetry_oban/plugin_handler.ex
```diff
@@ -3,64 +3,127 @@ defmodule OpentelemetryOban.PluginHandler do
...
 
-  def handle_plugin_start(_event, _measurements, %{plugin: plugin} = metadata, _config) do
-    OpentelemetryTelemetry.start_telemetry_span(
-      @tracer_id,
-      "#{plugin} process",
-      metadata,
-      %{}
-    )
+    attributes = %{
+      "messaging.oban.plugin": inspect(plugin)
```
indrekj, line 30: I don't understand why do we need inspect here

instrumentation/opentelemetry_oban/lib/opentelemetry_oban/plugin_handler.ex
```diff
@@ -3,64 +3,127 @@ defmodule OpentelemetryOban.PluginHandler do
...
+    }
+  end
+
+  defp plugin_work_attributes(%{plugin: Oban.Plugins.Gossip} = metadata) do
+    %{
+      "messaging.oban.gossip_count": metadata.gossip_count
+    }
+  end
+
+  defp plugin_work_attributes(%{plugin: Oban.Plugins.Reindexer}), do: %{}
```
indrekj, line 69: could also remove this line and let it go to the default case

instrumentation/opentelemetry_oban/lib/opentelemetry_oban/plugin_handler.ex
indrekj: Actually no. This is an atom right, so there wouldn't be extra quotes.

instrumentation/opentelemetry_oban/lib/opentelemetry_oban/plugin_handler.ex
indrekj: Got it. This is an atom right, so there wouldn't be extra quotes.

indrekj: Maybe `Atom.to_string()` or `"#{plugin}"` would be more intuitive though

indrekj (approved):
I didn't set up things to get it working on my computer and fully test it, but I went through the changes and they seem to make sense to me.

tsloughter:
@andrewhr any comments on @indrekj's notes?

yordis:
Superseded by #788, which fixes the underlying issues (including the JobHandler `:error` key bug flagged above) on a fresh branch. Closing in favor of that. Thanks for the original report, @andrewhr.

PR #198, opened 2023-08-23
Trace Liveview in Phoenix instrumentation

derekkraan:
Supercedes #108 since I don't have access to the source repo anymore.

I've removed the parent span that lives for the duration of the liveview process, and instead added `thread.name` with the pid, so that different events in the same liveview can be correlated to each other (per suggestion from @connorlay).

This sidesteps the whole discussion on monitoring spans. If we decide later that we want an overarching span, then that can always be added in at that point in time.

### DIFF
--- a/instrumentation/opentelemetry_phoenix/lib/opentelemetry_phoenix.ex
+++ b/instrumentation/opentelemetry_phoenix/lib/opentelemetry_phoenix.ex
@@ -10,6 +10,11 @@ defmodule OpentelemetryPhoenix do
                       default: nil,
                       doc: "The phoenix server adapter being used.",
                       type_doc: ":atom"
+                    ],
+                    liveview: [
+                      type: :boolean,
+                      default: true,
+                      doc: "Whether LiveView traces will be instrumented."
                     ]
                   )
 
@@ -78,6 +83,10 @@ defmodule OpentelemetryPhoenix do
     attach_router_start_handler()
     attach_router_dispatch_exception_handler()
 
+    if opts[:liveview] do
+      attach_liveview_handlers()
+    end
+
     :ok
   end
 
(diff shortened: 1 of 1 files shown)


tsloughter:
`thread.name` actually needs to be removed to merge this. That needs to be a general attribute added in the SDK and updated in the spec. I can handle that. If you remove it from here I can merge this.

derekkraan:
@tsloughter I'm not sure I understand, but I removed `thread.name` from the PR as requested.

tsloughter:
@derekkraan correct. tho needs to be configurable I assume since some SaaS charge for attributes.

tsloughter:
See https://github.com/open-telemetry/opentelemetry-erlang/pull/550/

It has `thread.name` as the scheduler id because I originally thought it would be that, but I came around to think pid was better.

tsloughter:
Looks like some formatting issues causing failure.

derekkraan:
I formatted with Elixir 1.15. I have seen the formatter checks fail between versions of Elixir, are you sure it's not that?

tsloughter:
Ah.

We need to change the formatting job to only run on 1 version.

tsloughter:
Opened https://github.com/open-telemetry/opentelemetry-erlang-contrib/issues/199 to track that

bryannaegele:
There are no tests for this functionality. Can you get those created in another PR @derekkraan?

tsloughter:
Yup, need those before any release.

connorlay:
Hey folks, I'm interested in getting this work published and am happy to help out with tests if needed @derekkraan @tsloughter 🙂

derekkraan:
I haven't been able to find time yet, so from my side it would be much appreciated!

skbolton:
> Yup, need those before any release.

Now that there is a PR with tests can a new release be cut with liveview tracing?

tielur:
> Yup, need those before any release.

I'd love to see a release with this included as well, I'm excited to try out telemetry with liveview

PR #200, opened 2023-08-26
Release drafting

bryannaegele:
Setting up monorepo release drafting to get tagging and automated release notes. We can maybe use this as a foundation for automating some other things when cutting releases since we'll have tags to work with in combo with the new tag-related rulesets.

Added missing test suites, labels, and re-alphabetized the actions file contents. I plan to refactor some of the test jobs to using shared workflows to make maintenance easier.

### DIFF
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -47,11 +47,32 @@ utilities:
 scope-ci:
   - .github/workflows/**
 
+opentelemetry_aws_xray:
+  - utilities/opentelemetry_aws_xray/**/*
+
 opentelemetry_cowboy:
   - instrumentation/opentelemetry_cowboy/**/*
 
+opentelemetry_dataloader:
+  - instrumentation/opentelemetry_dataloader/**/*
+
 opentelemetry_ecto:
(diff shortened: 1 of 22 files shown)


tsloughter:
ooo, i missed that you opened this. sweet.

tsloughter:
Are you down for github actions publishing to hex when the draft is turned into a full release?

tsloughter (approved):

bryannaegele:
I think we can get there with some more changes. For versioning at work, we're keeping a version file and reading that in like we used to do in some other libs.

Then we'd still need a manual workflow or something a generated PR that's locked down for only certain folks to be able to trigger the publishing. Add a publish dry-run step. All doable.

tsloughter:
Cool, yea. I think the publishing from draft should be enough to not require a manual workflow trigger but I'd be ok with it requiring a human to confirm the action after the dry run at which point it does the real publish.

bryannaegele:
I think last time I saw the topic of publishing from CI automatically in the maintainers channel it was shot down for security reasons amongst others.

PR #218, opened 2023-10-16
Create dependabot.yml

bryannaegele:
I'm gonna regret this but it's also the only way to stay on top of this for new releases.


tsloughter:
This is good but https://github.com/renovatebot/renovate does support Elixir and may be a better option.

tsloughter (approved):

bryannaegele:
Oh. It does show up when you search. I initially ruled it out because neither language was listed in "Language Support". I'll check it out

bryannaegele:
Ok. I'll look into renovate more since it appears we can do a custom manager to handle erlang

bryannaegele:
I'm going to close this in favor of getting renovate set up.

PR #221, opened 2023-11-01
Add tests to cover new liveview integration

derekkraan:
Following up on #198.


tsloughter (approved):
Thanks! I would like @bryannaegele to take a look before its merged, but I'm approving.

derekkraan:
Is there a reason that all the tests in the matrix are being skipped?

bryannaegele:
> Is there a reason that all the tests in the matrix are being skipped?

Forks can't trigger test runs without a maintainer giving it the initial trigger, I think. There's something there about that.

Given the number of tests we have and that the runners are shared with all of otel, I'm inclined to not change the behavior.

derekkraan:
Fixed the tests. Not sure why `mix format` ran with another version of Elixir when I submitted initially. This PR is ready for another look.

bryannaegele (approved):

PR #228, opened 2023-11-10
Update elixir-test-matrix.json

bryannaegele:
I'm going to thin this down to oldest and newest supported. The actions runs are getting a bit silly and this approach has worked fine in beam-telemetry for years.


bryannaegele:
The remaining failing erlang lib tests all appear to be from having out of date deps. Will address in separate PRs

PR #258, opened 2024-01-23
Finch: Sets the trace status to error on any 5xx response

jeffutter:
According to the [Semantic Conventions](https://opentelemetry.io/docs/specs/semconv/http/http-spans/#status) any trace for a 5xx response should mark the trace with `status: :error`. Previously they were not marked as error unless the client throws an error without reaching the server (timeout, connection refused, etc.)


instrumentation/opentelemetry_finch/lib/opentelemetry_finch.ex
```diff
@@ -67,6 +67,13 @@ defmodule OpentelemetryFinch do
         kind: :client
       })
 
+    if status >= 500 && status < 600 do
+      OpenTelemetry.Span.set_status(
+        s,
+        OpenTelemetry.status(:error, :opentelemetry_instrumentation_http.code_to_phrase(status))
```
bryannaegele, line 73: The reason cannot be set per convention. You can split the code to phrase to a separate PR or vice versa

utilities/opentelemetry_instrumentation_http/src/opentelemetry_instrumentation_http.erl
```diff
@@ -62,6 +65,201 @@ attribute_name(request, HeaderName) ->
...
+code_to_atom(504) -> gateway_timeout;
+code_to_atom(505) -> http_version_not_supported;
+code_to_atom(506) -> variant_also_negotiates;
+code_to_atom(507) -> insufficient_storage;
+code_to_atom(508) -> loop_detected;
+code_to_atom(510) -> not_extended;
+code_to_atom(511) -> network_authentication_required.
+
+-spec code_to_phrase(integer()) -> binary().
+code_to_phrase(100) -> <<"Continue">>;
```
bryannaegele, line 199: These values can all be atoms. Avoid strings for any value known at compile time. That removes the need for the `code_to_atom/1`.

utilities/opentelemetry_instrumentation_http/src/opentelemetry_instrumentation_http.erl
```diff
@@ -62,6 +65,201 @@ attribute_name(request, HeaderName) ->
 attribute_name(response, HeaderName) ->
     binary_to_atom(<<"http.response.header.", HeaderName/binary>>).
 
+-spec code(atom()) -> integer().
```
bryannaegele, line 68: What would be the use case for this?

instrumentation/opentelemetry_finch/lib/opentelemetry_finch.ex
jeffutter: Oh yeah I see
> Don’t set the span status description if the reason can be inferred from http.response.status_code.

That should remove the need for the util stuff altogether. 👍  I'll remove it

yordis:
Superseded by #470: on `main`, `set_span_status/3` sets `Error` for any status `>= 400`, which covers this. Thanks for raising it. The review discussion here is what #805 follows up on.

PR #259, opened 2024-01-24
HTTPoision: use atoms for span keys

jeffutter:
The other HTTP client implementations use atoms for span keys. Update the HTTPoison integration for consistency.

This shouldn't have any impact other than in testing.

This consistency is useful in that I'm working on a small suite of tests where I generate the same tests for each HTTP client and ensure they produce similar spans.


yordis:
Thanks for putting this together, and sorry this sat for so long.

I took another look against current `main`, and the repo now covers the same underlying goal through the newer semantic-conventions migration in #649.

Since this branch now conflicts and the current implementation/tests already use the newer attribute model, I’m going to close this as superseded rather than try to revive it here.

Thanks again for the contribution.

PR #260, opened 2024-01-24
Tesla: Include span attributes on error

jeffutter:
The call to `Tesla.run` could possibly return `{:error, _}` in certain cases (ex. timeout). Since attributes were set in a step after this, they were skipped in these scenarios.

This sets the attributes before the call to `Tesla.run` and then again after to cover the scenario that a middleware modified the attributes and they have been updated on the `Env`.

### DIFF
--- a/instrumentation/opentelemetry_tesla/lib/middleware/opentelemetry_tesla_middleware.ex
+++ b/instrumentation/opentelemetry_tesla/lib/middleware/opentelemetry_tesla_middleware.ex
@@ -33,8 +33,9 @@ defmodule Tesla.Middleware.OpenTelemetry do
       env
       |> maybe_put_additional_ok_statuses(opts[:mark_status_ok])
       |> maybe_propagate(Keyword.get(opts, :propagator, :opentelemetry.get_text_map_injector()))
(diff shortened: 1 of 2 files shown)


instrumentation/opentelemetry_tesla/lib/middleware/opentelemetry_tesla_middleware.ex
```diff
@@ -110,24 +117,37 @@ defmodule Tesla.Middleware.OpenTelemetry do
...
+  end
+
+  defp build_response_attrs(
+         %Tesla.Env{
+           status: status_code,
+           headers: headers
+         } = env
+       ) do
+    attrs =
+      Map.merge(build_base_attrs(env), %{
```
bryannaegele, line 148: What's the reasoning to replace all the attributes created during the request?

yordis:
Superseded by #610, which now sets request attributes (`http.request.method`, `server.address`, `server.port`, `url.full`) at span-start time, before `Tesla.run` executes. Verified attributes are present on adapter errors (including timeouts) on current main. Closing.

PR #262, opened 2024-01-31
Extra metadata to add to telemetry events for Redix

prabhjotlalli:
These end up in `:extra_metadata` metadata key to pass along.


bryannaegele:
Would folks always want all the metadata as attributes or just particular keys? I imagine folks would want more control.

PR #300, opened 2024-03-15
Pass attributes in context to enable child span for better sampling

azeemchauhan:
Discovered a use-case where the child span needs information from the parent span's attributes. I come up with putting the attributes in the current context and passing it around in the same trace and use it.


bryannaegele:
Hey @azeemchauhan. If you want to do something like this in phoenix then you could add a plug and put whatever you'd like into context there. For where you're looking to do this, you'd want your plug inserted into your endpoint right after the telemetry plug.

We generally cannot pollute the context in instrumentation as it has performance implications.

azeemchauhan:
@bryannaegele Thanks for the clarification. I was exploring options to get pass some information from parent span to the child span and then figured that there is no way of reading the parent's span attributes so I thought context could be right choice but looks like this will pollute the context and might have some performance implication due to extra info in context.


And, yes another way of achieving what I want is, update the Absinthe plug to add values from the `conn` ( ie. Host,  request headers etc) to the plug configuration and then pass to the Absinthe telemetry.

PR #303, opened 2024-03-18
Library to support Telemetry.Metrics as OpenTelemetry metrics

tsloughter:
Not ready because still using local changes in the deps but wanted to open it up for review while those changes are being merged and released.


tsloughter:
Forgot I needed to make `meter` a per-metric option. Though that isn't actually going to work since the main purpose of this is supporting stuff like Phoenix which isn't going to add meter to the reporter opts of telemetry events...

So may just add a warning in the docs that Scope is going to be wrong for all your `Telemetry.Metrics`

josevalim (approved):
😍

utilities/opentelemetry_telemetry_metrics/lib/otel_telemetry_metrics.ex
```diff
@@ -0,0 +1,184 @@
...
+      attach(meter, event_name, instruments)
+    end
+  end
+
+  defp create_instruments(meter, metrics) do
+    for metric <- metrics,
+      instrument = create_instrument(metric, meter, %{unit: unit(metric.unit)}),
+      instrument != nil, into: %{} do
+        {metric, instrument}
+    end
```
josevalim, line 91: Comprehensions automatically filter nil values, so you can skip `instrument != nil`.

```suggestion
    for metric <- metrics,
        instrument = create_instrument(metric, meter, %{unit: unit(metric.unit)}),
        into: %{} do
        {metric, instrument}
    end
```

utilities/opentelemetry_telemetry_metrics/lib/otel_telemetry_metrics.ex
```diff
@@ -0,0 +1,184 @@
...
+  Metrics of type `last_value` are ignored because `last_value` is not yet an
+  aggregation supported on synchronous instruments in Erlang/Elixir
+  OpenTelemetry. When it is added to the SDK this library will be updated to
+  no longer ignore metrics of this type.
+  """
+
+  require Logger
+  use GenServer
+
+  @doc """
```
josevalim, line 52:
```suggestion
  @doc """
  Starts open telemetry metrics reporter process.
  
  Accepts the same options as `GenServer.start_link/2`.
```

utilities/opentelemetry_telemetry_metrics/lib/otel_telemetry_metrics.ex
RudolfMan: umm.. I don't think it's true..

I just tried:
```elixir
iex> for x <- [1, nil, 2, 3], do: x
[1, nil, 2, 3]
```
neither it filters `nil` when passed `into: %{}`
```elixir
iex> for x <- [1, nil, 2, 3], into: %{}, do: {x, x}
%{1 => 1, 2 => 2, 3 => 3, nil => nil}
```

josevalim: Comprehension filters automatically do:

```
iex> for x <- [1, nil, 2, 3], x, into: %{}, do: {x, x}
```

RudolfMan: Oh, I see, thank you!

tsloughter: Thanks!

utilities/opentelemetry_telemetry_metrics/lib/otel_telemetry_metrics.ex
tsloughter: The last line seems wrong? Or at least confusing. That'd make me think the `options` are ones for configuring the `GenServer`, stuff like `hibernate`, but its the options to be passed to `init`.

tsloughter:
@AndrewDryga yes, definitely! Are you on slack? I hit a bug in metrics that has kept me from having time to spend on docs but could help you on slack to get going.

tsloughter:
WHOOPS. I started a job shortly after this and forgot about this PR, wow. Trying to resurrect it.

tsloughter:
Marked as ready for review tho I don't remember if it really is. But based on the tests it has I guess it works, hehe.

bryannaegele (approved):
Need to update the deps but let's get this in the hands of folks

utilities/opentelemetry_telemetry_metrics/mix.exs
```diff
@@ -0,0 +1,37 @@
...
+
+  # Run "mix help deps" to learn about dependencies.
+  defp deps do
+    [
+      {:telemetry, "~> 1.3"},
+      {:telemetry_metrics, "~> 1.0"},
+      {:opentelemetry_api, "~> 1.3"},
+      {:opentelemetry_api_experimental, "~> 0.5"},
+      {:opentelemetry, "~> 1.4.0"},
+      {:opentelemetry_experimental, "~> 0.5"},
```
bryannaegele, line 33:
```suggestion
      {:opentelemetry, "~> 1.4", only: [:dev, :test]},
      {:opentelemetry_experimental, "~> 0.5", only: [:dev, :test]},
```

ethangunderson:
I'm very interested in getting this merged. How can I help?

Fahani (approved):

tsloughter:
@ethangunderson  tried it by chance? That'd make it easier to feel confident merging.

ethangunderson:
@tsloughter I haven't yet, but I have the perfect app for it. I'll make time for it this week and report back.

ethangunderson:
Alright, a bit late, but I tried this out yesterday.

I have an existing application that uses the OTLP collectors from a Datadog agent. Traces work correctly, but I can't get metrics working. If I swap out the reader for `otel_metric_exporter_pid`, I do see the expected metrics in my console, so this is probably a DD config issue on my part. That being said, I'm not sure how to troubleshoot that any further. Do you have a metrics application that you use locally for verification of this stuff?

If the application does not have `opentelemetry_experimental` configured the Telemetry handler is detached with the following error. 

```
15:25:07.397 [error] Handler {OtelTelemetryMetrics, [:telemetry, :event], #PID<0.935.0>} has failed and has been detached. Class=:error
Reason=:undef
Stacktrace=[
  {:otel_meter_noop, :record,
   [
     %{
       {:otel_tracer, :span_ctx} => {:span_ctx,
        260427163911476008642819364899241647806, 15918744878643449198, 1,
        {:tracestate, []}, true, false, true,
        {:otel_span_ets,
         #Function<2.20827891/1 in :otel_tracer_server.on_end/1>}}
     },
```

tsloughter:
@ethangunderson are you sending the metrics to the collector which then sends to DataDog? Can you verify if they are getting to the cluster by adding the debug exporter to the collector metrics pipeline.

zkayser:
@tsloughter I work with @ethangunderson and just got a little test using this working. We're actually sending metrics directly to a Datadog Agent over grpc on port 4317 and I got it to work fine, just a few things that tripped me up along the way. I'll provide the three main things I ran into (I hacked a couple of them into working just to prove this out as a PoC).

First, here's the config we're using for reference

```elixir
config :opentelemetry_experimental,
  readers: [
    %{
      module: :otel_metric_reader,
      config: %{
        exporter:
          {:otel_exporter_metrics_otlp,
           %{
             endpoint: System.get_env("OTLP_ENDPOINT"),
             protocol: :grpc,
             compression: :gzip
           }}
      }
    }
  ]

```

We set up a little test with a single `Telemetry.Metrics.counter/2 ` metric:

```elixir
defmodule Vantage.Otel.Metrics do
  @moduledoc """
  Experimental usage of OtelTelemetryMetrics for a PoC
  """
  use Supervisor

  import Telemetry.Metrics

  def start_link(opts) do
    Supervisor.start_link(__MODULE__, opts, name: __MODULE__)
  end

  @impl Supervisor
  def init(opts) do
    children = [
      {OtelTelemetryMetrics, metrics: Keyword.get(opts, :metrics, metrics())}
    ]

    Supervisor.init(children, strategy: :one_for_one)
  end

  defp metrics do
    [
      counter("vantage.test.metric.count",
        # unit: :count, ====> Needed to set this explicitly to avoid an error in `otel_otlp_metrics.erl`
        event_name: [:vantage, :test, :metric],
        description: "just a test",
        measurement: 1
      )
    ]
  end
end
```

### 1) ArgumentError on `:unicode.characters_to_binary/1` without explicitly passing a value to `:unit` on `Telemetry.Metrics` opts

Without explicitly setting a value for `:unit` in the `Telemetry.Metrics` options, we hit this exception:

```
10:17:01.357 [error] GenServer #PID<0.428.0> terminating
** (ArgumentError) errors were found at the given arguments:

  * 1st argument: not valid character data (an iodata term)

    (stdlib 6.1.2) unicode.erl:1219: :unicode.characters_to_binary(1)
    (opentelemetry_experimental 0.5.1) /Users/zkayser/dev/carburetor/deps/opentelemetry_experimental/apps/opentelemetry_experimental/src/otel_otlp_metrics.erl:72: :otel_otlp_metrics.to_proto/1
```
 
 We were able to get around this by passing a value under the `:unit` option on the `Telemetry.Metrics` definition.
 
 ### 2) Undefined instrumentation scope on default meter
 
This may be user error on my part or we should have been using our own `#meter{}`, but with the default `meter` we ran into this exception:

```
10:29:09.651 [error] GenServer #PID<0.493.0> terminating
** (stop) {:badrecord, :undefined}
    (opentelemetry_experimental 0.5.1) /Users/zkayser/dev/carburetor/deps/opentelemetry_experimental/apps/opentelemetry_experimental/src/otel_otlp_metrics.erl:51: anonymous fn/3 in :otel_otlp_metrics.to_proto_by_scope/1
```

I hacked this one a bit to get around this issue. This may not be an issue if there is something similar to `create_application_tracers/1` but for `meters` instead. I hacked this by adding `instrumentation_scope = opentelemetry:instrumentation_scope(<<>>, <<>>, <<>>)` to the default meter in `otel_meter_server.erl`.

3) Undefined `description` on metric caused error on generated metric service protobuf module

The last error I ran into was that without a `#metric` record having its `description` field set, it failed in `opentelemetry_exporter_metrics_service_pb.erl` due to a function clause missing error:

```
10:35:37.249 [error] GenServer #PID<0.468.0> terminating
** (FunctionClauseError) no function clause matching in :opentelemetry_exporter_metrics_service_pb.is_empty_string/1
    (opentelemetry_exporter 1.8.0) /Users/zkayser/dev/carburetor/deps/opentelemetry_exporter/src/opentelemetry_exporter_metrics_service_pb.erl:2171: :opentelemetry_exporter_metrics_service_pb.is_empty_string(:undefined)
```


There is not much in this feedback that pertains specifically to this PR other than maybe getting tripped up on the `:unit` value. I admit that we are just now getting started with Otel Metrics (we've been reporting metrics over Statsd for years now) and that there may indeed be some user error on our part. 

If there is anything at all I can do to help make this feedback more actionable, or if there are any issues I can work on, I would be happy to help out in whatever way we can!

cheerfulstoic:
Also important, here is my configuration! It took me quite a while to figure it out, so hopefully it's useful!

The "cumulative" settings under `default_temporality_mapping` are there because it seems like prometheus needs them. So anybody following along may not want those.

```elixir
  config :opentelemetry,
    span_processor: :batch,
    traces_exporter: :otlp

  config :opentelemetry_exporter,
    otlp_protocol: :http_protobuf,
    otlp_endpoint: opentelemetry_endpoint

  config :opentelemetry_experimental,
    readers: [
      %{
        module: :otel_metric_reader,
        config: %{
          export_interval_ms: 1_000,
          exporter: {
            :otel_exporter_metrics_otlp,
            %{endpoints: [opentelemetry_endpoint], protocol: :http_protobuf, ssl_options: []}
          },
          # For Prometheus
          default_temporality_mapping: %{
            counter: :temporality_cumulative,
            observable_counter: :temporality_cumulative,
            updown_counter: :temporality_cumulative,
            observable_updowncounter: :temporality_cumulative,
            histogram: :temporality_cumulative,
            observable_gauge: :temporality_cumulative
          }
        }
      }
    ]
```

cheerfulstoic:
👍  Cool

Yeah, I was using the `grafana/otel-lgtm` container to get the metrics and when I turned on `ENABLE_LOGS_PROMETHEUS` I was getting errors about the metrics until I changed them to cumulative.  So I left them that way because we use prometheus in production as well, though we're still working to understand how the metrics are coming out into grafana.

utilities/opentelemetry_telemetry_metrics/lib/otel_telemetry_metrics.ex
```diff
@@ -0,0 +1,189 @@
...
+
+  defp detach(handler_ids) do
+    Enum.each(handler_ids, fn id -> :telemetry.detach(id) end)
+  end
+
+  defp handler_id(event_name) do
+    {__MODULE__, event_name, self()}
+  end
+
+  def handle_event(_event_name, measurements, metadata, %{meter: meter,
```
yordis, line 150: Wouldn't this cause a OOMKill because of the gen server mailbox?

tsloughter: How so?

yordis: ignore me 🤦🏻 sorry... I just confused with a gen server handling ... my bad!

Issue #308, opened 2024-04-05
OpentelemetryBandit is incompatible with Bandit v1.4.0


ibarchenkov:
**Describe the bug**
In Bandit 1.4.0, `status`, `method` and `request_target` attributes [were removed](https://github.com/mtrudel/bandit/blob/main/CHANGELOG.md#changes-1) from telemetry metadata, which are utilised in [multiple places](https://github.com/open-telemetry/opentelemetry-erlang-contrib/blob/3f13bc8b2042071b723d0ea8c7b118fd10f42d26/instrumentation/opentelemetry_bandit/lib/opentelemetry_bandit.ex#L68) in OpentelemetryBandit.

**Expected behavior**
These attributes should be replaced with corresponding data from the `conn` struct.

**Additional context**
I'm willing to provide a PR if no one is working on this issue yet.

VictorGaiva:
There is an [open PR](https://github.com/samokat-oss/opentelemetry-bandit/pull/1) for fixing this issue on the original repository for the package.

PR #310, opened 2024-04-10
Support Bandit 1.4 telemetry changes

elvanja:
Adjusts to the telemetry breaking changes in Bandit 1.4.

See https://github.com/mtrudel/bandit/blob/main/CHANGELOG.md#140-26-mar-2024
Related to https://github.com/open-telemetry/opentelemetry-erlang-contrib/issues/308.


instrumentation/opentelemetry_bandit/lib/opentelemetry_bandit.ex
```diff
@@ -44,6 +44,7 @@ defmodule OpentelemetryBandit do
     duration = measurements.duration
     end_time = :opentelemetry.timestamp()
     start_time = end_time - duration
+    resp_body_bytes = Map.get(measurements, :resp_body_bytes, 0)
```
elvanja, line 47: This is because in another app using Bandit 1.4 and this fix, encountered a situation where `:resp_body_bytes` was not set at all in measurements.

instrumentation/opentelemetry_bandit/mix.exs
```diff
@@ -74,7 +73,7 @@ defmodule OpentelemetryBandit.MixProject do
       # dev dependencies
       {:ex_doc, ">= 0.0.0", only: :dev, runtime: false},
       {:excoveralls, "~> 0.18", only: :test},
-      {:bandit, "~> 1.0", only: [:dev, :test], runtime: false},
+      {:bandit, "~> 1.4", only: [:dev, :test], runtime: false},
```
elvanja, line 76: The idea being that at least 1.4 is needed for this. Not sure if it would make sense to keep 1.0 dependency and make code a bit smarter, e.g. find `:method` in `meta` or `conn`.

Issue #136, opened 2022-11-25
Thousand Island instrumentation


bryannaegele:
With Bandit now having full support for being a drop-in cowboy replacement for Phoenix, we'll need an instrumentation library for Thousand Island from [its telemetry](https://hexdocs.pm/thousand_island/ThousandIsland.html#module-logging-telemetry).

mtrudel:
Note that thousand island's telemetry model is being overhauled to better accommodate exactly this scenario. See provisional work at https://github.com/mtrudel/thousand_island/pull/27 (which itself is largely blocked by knowing what the golden path is for exactly this integration).

tsloughter:
I'm not sure how instrumentation of the socket level can actually be correlated with the application level.

Any span created in ThousandIsland will be ignored when the HTTP request's propagated context is extracted.

We'll want to create metrics from these events, of course, but I don't think they can be used for spans.

mtrudel:
My *really* early take on this at the HTTP layer is up at https://github.com/mtrudel/bandit/tree/telemetry; generally I was planning to

1. Have Bandit reach into the underlying ThousandIsland.Socket to get the connection span to use as a parent:
2. Have Bandit create spans / events as appropriate for each HTTP request & the various phases of a request (header & body reading, response writing, etc).
3. Put the relevant span into the `Plug.Conn` struct for a request (likely as a `private`) and somehow get phoenix to nest its spans within this if present.

Truthfully, I'm not at all confident I'm not overbuilding this. The time lookups alone are pretty expensive to be doing so many of them.

tsloughter:
> Have Bandit reach into the underlying ThousandIsland.Socket to get the connection span to use as a parent:

Yea, that is an issue since Bandit, or Plug/Phoenix, needs to create a span from a parent that is received in request headers. It'll have its own trace id, so can't have a parent/child relationship with anything from ThousandIsland.

The only option I know of for such a relationship would be that Bandit could add the ThousandIsland span as a linked span if it is what starts the span from the propagated headers.

mtrudel:
Interesting! Alright, so if I understand correctly:

* Since all the trade id header work is in open telemetry's Phoenix support, there's nothing for Bandit / Thousand Island to do in order for phoenix's telemetry support to 'just work', including support for passed in parent traces

[...]

PR #311, opened 2024-04-10
Support Bandit 1.4 telemetry changes

elvanja:
Adjusts to the telemetry breaking changes in Bandit 1.4.

See https://github.com/mtrudel/bandit/blob/main/CHANGELOG.md#140-26-mar-2024
Related to https://github.com/open-telemetry/opentelemetry-erlang-contrib/issues/308.


instrumentation/opentelemetry_bandit/mix.exs
```diff
@@ -74,7 +73,7 @@ defmodule OpentelemetryBandit.MixProject do
       # dev dependencies
       {:ex_doc, ">= 0.0.0", only: :dev, runtime: false},
       {:excoveralls, "~> 0.18", only: :test},
-      {:bandit, "~> 1.0", only: [:dev, :test], runtime: false},
+      {:bandit, "~> 1.4", only: [:dev, :test], runtime: false},
```
elvanja, line 76: The idea being that at least 1.4 is needed for this. Not sure if it would make sense to keep 1.0 dependency and make code a bit smarter, e.g. find `:method` in `meta` or `conn`.

instrumentation/opentelemetry_bandit/lib/opentelemetry_bandit.ex
```diff
@@ -44,6 +44,7 @@ defmodule OpentelemetryBandit do
     duration = measurements.duration
     end_time = :opentelemetry.timestamp()
     start_time = end_time - duration
+    resp_body_bytes = Map.get(measurements, :resp_body_bytes, 0)
```
elvanja, line 47: This is because in another app using Bandit 1.4 and this fix, encountered a situation where :resp_body_bytes was not set at all in measurements.

danschultzer: Do you know in what case this happens? Might be good to reproduce in the test. I was trying to understand the code path in Bandit but from how I read it the `:resp_body_bytes` should always be set.

elvanja: So it seems that this occurs when the route does not exist. 
In my local setup, have a Prometheus service which tries to hit `GET /metrics` which does not exist.
And then the meta, conn and measurements look like this, note that indeed `:resp_body_bytes` is not in measurements:

```elixir
# >>> meta
%{
  conn: %Plug.Conn{
    adapter: {Bandit.Adapter, :...},
    assigns: %{},
    body_params: %Plug.Conn.Unfetched{aspect: :body_params},
    cookies: %Plug.Conn.Unfetched{aspect: :cookies},
    halted: false,
    host: "host.docker.internal",
    method: "GET",
    owner: #PID<0.1802.0>,
    params: %Plug.Conn.Unfetched{aspect: :params},
    path_info: ["metrics"],
    path_params: %{},
    port: 4000,
    private: %{phoenix_endpoint: CustodyWeb.Endpoint},
    query_params: %Plug.Conn.Unfetched{aspect: :query_params},
    query_string: "",
    remote_ip: {127, 0, 0, 1},
    req_cookies: %Plug.Conn.Unfetched{aspect: :cookies},
    req_headers: [
      {"x-prometheus-scrape-timeout-seconds", "5"},
      {"accept-encoding", "gzip"},
      {"accept",
       "application/openmetrics-text;version=1.0.0,application/openmetrics-text;version=0.0.1;q=0.75,text/plain;version=0.0.4;q=0.5,*/*;q=0.1"},
      {"user-agent", "Prometheus/2.37.9"},
      {"host", "host.docker.internal:4000"}
    ],
    request_path: "/metrics",
    resp_body: nil,
    resp_cookies: %{},
    resp_headers: [{"cache-control", "max-age=0, private, must-revalidate"}],
    scheme: :http,
    script_name: [],
    secret_key_base: :...,
    state: :sent,
    status: nil
  },
  telemetry_span_context: #Reference<0.3530071607.2258632705.218907>,
  connection_telemetry_span_context: #Reference<0.3530071607.2258632705.218854>
}

# >>> measurements:
%{
  monotonic_time: -576460731151783373,
  duration: 243141570,
  req_header_end_time: -576460731375909890
}
```

elvanja: As far as i can tell, the https://github.com/elvanja/opentelemetry-erlang-contrib/blob/6a3ad343af8211706427f4deb17d1570ed87a9ff/instrumentation/opentelemetry_bandit/test/bandit/opentelemetry_bandit_test.exs#L98-L127 test checks exactly that. But, it also sets `:resp_body_bytes` so the tests doesn't fail even without the fix.

Tried checking Bandit flow too and as per https://github.com/mtrudel/bandit/blob/4f15f029e7aa17f8e7f98d55b0e94c684dee0971/lib/bandit/adapter.ex#L237, it seems that the measurements should contain `:resp_body_bytes` at all times.

elvanja: One more thing. Bandit adapter contains `:resp_body_bytes` in metrics. This is from [Adapter.send_resp](https://github.com/mtrudel/bandit/blob/4f15f029e7aa17f8e7f98d55b0e94c684dee0971/lib/bandit/adapter.ex#L140):
```elixir
%Bandit.Adapter{
  transport: %Bandit.HTTP1.Socket{
    socket: %ThousandIsland.Socket{
      socket: #Port<0.27>,
      transport_module: ThousandIsland.Transports.TCP,
      read_timeout: 60000,
      silent_terminate_on_error: false,
      span: %ThousandIsland.Telemetry{
        span_name: :connection,
        telemetry_span_context: #Reference<0.3423108854.2804940801.55897>,
        start_time: -576460742183292227,
        start_metadata: %{
          remote_address: {127, 0, 0, 1},
          remote_port: 55941,
          telemetry_span_context: #Reference<0.3423108854.2804940801.55897>,
          parent_telemetry_span_context: #Reference<0.3423108854.2804940804.49853>
        }
      }
    },
    buffer: "",
    read_state: :read,
    write_state: :sent,
    bytes_remaining: nil,
    body_encoding: nil,
    version: :"HTTP/1.1",
    send_buffer: [],
    keepalive: true,
    opts: %{http: [], websocket: [], http_1: [], http_2: []}
  },
  owner_pid: #PID<0.1726.0>,
  method: "GET",
  status: 404,
  content_encoding: "gzip",
  upgrade: nil,
  metrics: %{
    req_header_end_time: -576460742178858184,
    resp_uncompressed_body_bytes: 75227,
    resp_compression_method: "gzip",
    resp_start_time: -576460742006448971,
    resp_body_bytes: 13034,
    resp_end_time: -576460742004810980
  },
  opts: %{http: [], websocket: [], http_1: [], http_2: []}
}
```

However, in [Telemetry.stop_span](https://github.com/mtrudel/bandit/blob/4f15f029e7aa17f8e7f98d55b0e94c684dee0971/lib/bandit/telemetry.ex#L188), `:resp_body_bytes` is not in measurements:
```elixir
%{
   monotonic_time: -576460741994242499,
   req_header_end_time: -576460742178858184,
   duration: 188840147
 }
```

elvanja: If it helps, tracing the flow in https://github.com/mtrudel/bandit/blob/4f15f029e7aa17f8e7f98d55b0e94c684dee0971/lib/bandit/pipeline.ex#L35-L41, it seems that `conn.adapter.metrics` in each step do not have any metrics generated in https://github.com/mtrudel/bandit/blob/4f15f029e7aa17f8e7f98d55b0e94c684dee0971/lib/bandit/adapter.ex#L140. I see metrics generated in `Bandit.Adapter`, but in `Bandit.Telemetry` lands adapter struct without those metrics.

danschultzer: Thanks! I didn't grok what happens on 404s here, but looking up all `Bandit.Telemetry.stop_span` calls there are a two places where the metrics are not added but the `:stop` telemetry event is send:

https://github.com/mtrudel/bandit/blob/4f15f029e7aa17f8e7f98d55b0e94c684dee0971/lib/bandit/pipeline.ex#L54
https://github.com/mtrudel/bandit/blob/4f15f029e7aa17f8e7f98d55b0e94c684dee0971/lib/bandit/pipeline.ex#L74

`Bandit.Telemetry` docs does explicitly state that the `:resp_body_bytes` measurement for `[:bandit, :request, :stop]` events is set, so I think this is an issue upstream in bandit. I've opened an issue: https://github.com/mtrudel/bandit/issues/340

elvanja: I get the idea but it seems that this is not an error state after all. E.g. the `stop_span` occurs in https://github.com/mtrudel/bandit/blob/4f15f029e7aa17f8e7f98d55b0e94c684dee0971/lib/bandit/pipeline.ex#L41 for this situation. It does not go to line#54 or line#74, at least that is how it behaves locally for me.

instrumentation/opentelemetry_bandit/mix.exs
bryannaegele: I'm going to start on the refactor with the bandit event changes and http semantic conventions which will be a breaking change in itself. 

This PR is really helpful in identifying some of the issues y'all have found with the update. We might need some further tweaks by Mat. Can you aggregate what you've found thus far into #136?

michelboaventura:
Hey. Is there any help needed here to merge this PR that we can do? We're not able to upgrade bandit because of this issue.

bryannaegele:
Thank you for all this. Big help.

I will be merging this and #317 so I can knock out any remaining functionality changes needed and the new semcon. There will be some changes before the new release can be made so please consider everything WIP still.

bryannaegele (approved):

PR #313, opened 2024-04-12
Allow to customize span names in `opentelemetry_ecto`

whatyouhide:
`my_app.repo.query:users` is sometimes not a good-enough name, I think. Unless I’m missing something, sometimes you might want a custom name, such as `my_app.repo.get_all_users` or `my_app.repo.insert_post` or something like that. For example, we attach info like that to queries via Ecto's `:telemetry_options`:

```elixir
MyApp.Repo.all(User, telemetry_options: [source: "my_app.repo.get_all_users"])
```

Thoughts on the solution?

### DIFF
--- a/instrumentation/opentelemetry_ecto/lib/opentelemetry_ecto.ex
+++ b/instrumentation/opentelemetry_ecto/lib/opentelemetry_ecto.ex
@@ -29,6 +29,7 @@ defmodule OpentelemetryEcto do
           | {:span_prefix, String.t()}
           | {:additional_attributes, %{String.t() => term()}}
           | {:db_statement, :enabled | :disabled | (String.t() -> String.t())}
+          | {:span_name, (:telemetry.event_metadata() -> String.t())}
 
   @doc """
   Attaches the `OpentelemetryEcto` handler to your repo events.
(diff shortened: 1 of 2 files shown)


tsloughter:
I'm not sure about ignoring span prefix.

whatyouhide:
@tsloughter you think we should go with `:span_prefix` prefixing whatever is returned by `:span_name`? I’m ok with that.

bryannaegele:
I can't find any provision for allowing this in the conventions, but they're always changing. https://opentelemetry.io/docs/specs/semconv/database/database-spans/

whatyouhide:
> I do wish we had better naming here but honestly this should all be down in postgrex or whatever

Why? Genuine question as I’m really new to OTel. Ecto seemed like a great place due to the fact that the OTel instrumentation works regardless of adapter, so I’m curious to understand why Postgrex would be a better place for the instrumentation.

whatyouhide:
Okay no I read through the semantic conventions and I have better ideas. I'll be back™

bryannaegele:
> Why? Genuine question as I’m really new to OTel
We're really kinda limited to what's available in the ecto telemetry and it's not a lot. Ecto is generic so anything someone would want from the adapter isn't there since it's not there. A perfect example is every operation for `ayesql` ends up being `prefix:query` and you have to parse/key in on the statement to try to discern things in the rest of your tooling.

> I’m curious to understand why Postgrex would be a better place for the instrumentation.
I'm gonna get shot for saying it, but Ecto is essentially an ORM and thus a level higher than where DB or other store instrumentation is desired to be tracked (otel desire, not mine). For .NET, it would be equivalent of only instrumenting in Entity Framework or Hibernate in Java. I know it's not 1:1 but they built the spec to be at the adapter level. Anything above that layer would be separate instrumentation.

If we had instrumentation hooks into postgrex and the other adapters, we'd move that instrumentation to there. Similar to how we use cowboy instrumentation for Phoenix and then decorate anything else because cowboy is the thing that's doing the real work and is at the real boundary of the request/response.

whatyouhide:
> First stop for everything we do in anything :)

This was definitely my TIL for today 🙃 

All you said makes sense, thanks for the context!

PR #316, opened 2024-04-22
Add opentelemetry_xandra package

whatyouhide:
This should be pretty useful since Xandra exposes low-level Telemetry events. @bryannaegele mentioned that this approach should be better than what `opentelemetry_ecto` does since it gives us access to the database-level events instead of the ORM-level events.

The library is pretty simple and is mostly patched together from other libs in this repo. Thoughts? 🙃


tsloughter (approved):
This will need to be updated to use the new semconv when it is released but I think can just hardcode as it does now until then bc they'll just have to change anyway.

PR #317, opened 2024-04-24
Bandit: Start span when connection start rather than retroactively

shadowfacts:
This ensures that spans created in application code are correctly parented to the request span, like when using the Cowboy or default Phoenix handlers.

This PR builds on top of @elvanja's work in #311.

TODOs:
- [ ] Actually test the websocket changes (the project I've been testing with has no websockets)


instrumentation/opentelemetry_phoenix/lib/opentelemetry_phoenix.ex
```diff
@@ -173,11 +173,6 @@ defmodule OpentelemetryPhoenix do
     |> OpenTelemetry.Ctx.attach()
   end
 
-  defp bandit_start() do
-    OpentelemetryProcessPropagator.fetch_parent_ctx()
-    |> OpenTelemetry.Ctx.attach()
-  end
```
shadowfacts, line 179: This function seems to have just been copied from the Cowboy adapter, but it breaks propagating the context for Bandit. This is because the telemetry events for Bandit are fired in the same process that handles the request (i.e., the same process that fires the Phoenix telemetry events). In effect, this was unsetting the already-set context.

khustochka:
This is a good change and would be nice if it were merged.

khustochka:
Websocket changes work fine. The only problem is that websocket connection can be open indefinitely long, and I do not know if there are any use of extra-long spans. I exports traces to Datadog, and it considers a span closed after 10 minutes. Also, if I am not mistaken, they have no child spans. I am not sure if cowboy handler traces websocket events.

bryannaegele:
> Websocket changes work fine. The only problem is that websocket connection can be open indefinitely long, and I do not know if there is any use of extra-long spans. I export traces to Datadog, and it considers a span closed after 10 minutes. Also, if I am not mistaken, they have no child spans and not a lot of useful attributes. I am not sure if cowboy handler creates websocket traces.

Websocket traces for the duration of a connection are not a thing. The recommended approach for folks if they want to correlate socket messages to a connection is to create a session id in the socket handler and add that to your spans.

There's some other good stuff in here. I have very limited time to get everything out so I will likely merge this along with the other PR and remove the sockets bit.

bryannaegele (approved):
Websocket functionality will get dropped but everything else is a great start. Thank you!

bryannaegele:
@shadowfacts can you sign the CLA ASAP please? We can't accept your work without it.

bryannaegele (commented):
Tests broken.

shadowfacts:
@bryannaegele Sorry for the delay. I'm entirely happy to grant copyright/patent licenses to the project and to assign any other rights to this work, but the CLA asks for my full legal name and my mailing address, which I'm not willing to provide.

bryannaegele:
No worries. We've asked that the CLA get less intrusive.

hyphenPaul:
@bryannaegele Is this work abandoned? This feature would be very useful for my team. We're currently using OTel with cowboy, but would like to switch over to bandit.

bryannaegele:
No, we just can't accept non-cla code. The spirit of what this was trying to do is covered in #357 though, so it won't be long

PR #318, opened 2024-04-24
chore(oban): allow passing options to insert wrapper

Miradorn:
Tried to use the wrapper and ran into problems since we use the optional `opts` argument in some places, which the wrapper currently doesn't support.
This will fix it.


Miradorn:
@yordis sorry for the direct ping, but since this is a pretty old PR that I just updated I wasn't sure if anybody will see it and it looks like you might be the most applicable person to tag. thanks for your work on this!

yordis (approved):

PR #152, opened 2023-02-28
Add ExAws instrumentation

Cantido:
[ExAws](https://github.com/ex-aws/ex_aws) just merged the necessary telemetry changes that include the AWS service name and operation in the telemetry metadata. It hasn't been included in a published version yet, but I wanted to get this ready for when it does, since my company uses ExAws heavily.

PR #330, opened 2024-06-05
Ex aws instrumentation

prithvihv:
https://github.com/open-telemetry/opentelemetry-erlang-contrib/pull/152


instrumentation/opentelemetry_ex_aws/lib/opentelemetry_ex_aws/dynamo_db.ex
```diff
@@ -0,0 +1,180 @@
...
+
+  defp request_attributes(%{request_body: request_body}, attribute_definition) do
+    extract_attributes(request_body, Map.get(attribute_definition, :request, []))
+  end
+
+  defp response_attributes(%{response_body: response_body}, attribute_definition) do
+    extract_attributes(response_body, Map.get(attribute_definition, :response, []))
+  end
+
+  defp response_attributes(%{result: :error}, _), do: %{}
```
mrnovalles, line 156: shouldn't there be a `Tracer.set_status()` call based on the error result?
As an example in `req`:
https://github.com/open-telemetry/opentelemetry-erlang-contrib/blob/main/instrumentation/opentelemetry_req/lib/opentelemetry_req.ex#L108

tsloughter: I think it depends. An error response doesn't mean what the Span is doing errored, just that the response from the backend was an error, if that makes sense...

tsloughter:
Thanks @prithvihv !

instrumentation/opentelemetry_ex_aws/lib/opentelemetry_ex_aws/dynamo_db.ex
prithvihv-wooga: Maybe we should set error, 

The tesla middleware assumes the span is an error if the response is 4xx.

yordis:
@prithvihv @prithvihv-wooga are you still around to get this to the finish line?

PR #332, opened 2024-06-14
feat: allow to pass repo name to setup

yordis:


instrumentation/opentelemetry_ecto/lib/opentelemetry_ecto.ex
```diff
@@ -38,13 +38,13 @@ defmodule OpentelemetryEcto do
...
   By default, it's the snake-cased name of the repository module. For `MyApp.Repo`, it would
-  be `[:my_app, :repo]`.
+  be `[:my_app, :repo]`. Or you could pass the repo module.
 
   For example:
 
       @impl Application
       def start(_type, _args) do
-        OpentelemetryEcto.setup([:blog, :repo])
+        OpentelemetryEcto.setup(Blog.Repo)
```
yordis, line 47: Encourage people to avoid making mistake with this one, hopefully.

instrumentation/opentelemetry_ecto/lib/opentelemetry_ecto.ex
```diff
@@ -240,4 +242,14 @@ defmodule OpentelemetryEcto do
...
   end
+
+  defp telemetry_prefix(event_prefix) when is_list(event_prefix) do
+    event_prefix
+  end
+
+  defp telemetry_prefix(repo) when is_atom(repo) do
+    repo
+    |> Kernel.apply(:config, [])
+    |> Keyword.get(:telemetry_prefix)
```
GregMefford, line 253: I think my main concern here would be if this will always return the right thing, or does it return `nil` in the case where a user hasn't overridden the default `telemetry_prefix` setting. I didn't look at the Ecto code, but the docs don't seem to indicate what fields will be there in the config - it's just whatever you configured in the app config or added in the `init/2` callback at runtime.

GregMefford (commented):
I don't have major concerns about this if we can implement it in a reliable way, but I haven't confirmed yet that the proposed change is robust.

instrumentation/opentelemetry_ecto/lib/opentelemetry_ecto.ex
yordis: What are the actionable items you would like me to take here?

bryannaegele: > What are the actionable items you would like me to take here?

I think he's asking how are you going to safeguard where the config doesn't have `telemetry_prefix` defined? This presumes it's there so if it isn't, it will return `nil`.

instrumentation/opentelemetry_ecto/lib/opentelemetry_ecto.ex
```diff
@@ -240,4 +242,14 @@ defmodule OpentelemetryEcto do
...
     Map.merge(attributes, additional_attributes)
   end
+
+  defp telemetry_prefix(event_prefix) when is_list(event_prefix) do
+    event_prefix
+  end
+
+  defp telemetry_prefix(repo) when is_atom(repo) do
+    repo
+    |> Kernel.apply(:config, [])
```
whatyouhide, line 252: If you call `repo.config([])` here instead it'll already make it easier for the Elixir type system to detect if `repo` is a module and so on.

yordis: It's one of those things that you do on Friday, and when you come back on Monday, you ask, "What was I thinking here?"

Thank you 💜

instrumentation/opentelemetry_ecto/lib/opentelemetry_ecto.ex
yordis: Gathering notes over here,

Based on the docs, it seems that, at the very least, there is a default value: https://github.com/elixir-ecto/ecto/blob/c97c9bad699a0349d06877cef2ee0f12495eaee9/lib/ecto/repo.ex#L49-L57

yordis: Based on https://github.com/elixir-ecto/ecto/blob/c97c9bad699a0349d06877cef2ee0f12495eaee9/lib/ecto/repo/supervisor.ex#L23-L28

There is still a possibility that somebody messes up the `repo_init`, so the question is, should we do defensive programming here or not?

I fixed the situation anyway

yordis:
Hey team, check the latest changes! I addressed your concerns

instrumentation/opentelemetry_ecto/lib/opentelemetry_ecto.ex
```diff
@@ -38,13 +38,13 @@ defmodule OpentelemetryEcto do
 
   `event_prefix` must be the prefix configured in the `Ecto.Repo` Telemetry configuration.
   By default, it's the camel_case name of the repository module. For `MyApp.Repo`, it would
-  be `[:my_app, :repo]`.
+  be `[:my_app, :repo]`. Or you could pass the repo module.
```
whatyouhide, line 41: Can you rephrase this to this:

```
`event_prefix` must be either the name of the Ecto repo, or the prefix configured in
the repo's Telemetry configuration. By default, it's the camel-cased name of the
repository module. For `MyApp.Repo`, for example, the default prefix would be
`[:my_app, :repo]`.
```

instrumentation/opentelemetry_ecto/lib/opentelemetry_ecto.ex
```diff
@@ -72,9 +72,10 @@ defmodule OpentelemetryEcto do
       query statements will not be recorded on spans.
 
   """
-  @spec setup(:telemetry.event_name(), [setup_option()]) :: :ok | {:error, :already_exists}
+  @spec setup(event_prefix :: module() | :telemetry.event_name(), [setup_option()]) ::
```
whatyouhide, line 75:
```suggestion
  @spec setup(module() | :telemetry.event_name(), [setup_option()]) ::
```

whatyouhide: Also, you can replace `module()` with `Ecto.Repo.t()`

whatyouhide: Aaaand one last thing, can we rename `event_prefix` to `repo_or_event_prefix` in docs and code?

instrumentation/opentelemetry_ecto/lib/opentelemetry_ecto.ex
```diff
@@ -240,4 +241,15 @@ defmodule OpentelemetryEcto do
...
   end
+
+  defp telemetry_prefix(event_prefix) when is_list(event_prefix) do
+    event_prefix
+  end
+
+  defp telemetry_prefix(repo) when is_atom(repo) do
+    case Keyword.get(repo.config(), :telemetry_prefix) do
+      telemetry_prefix when is_list(telemetry_prefix) -> telemetry_prefix
+      _ -> raise ArgumentError, ":telemetry_prefix not set in #{inspect(repo)} repo config"
```
whatyouhide, line 252: Do we need a test for this (and passing the module as well)?

whatyouhide (commented):
Left a couple more comments to address before merging 🙃

yordis:
@whatyouhide, it is ready to go now!

instrumentation/opentelemetry_ecto/lib/opentelemetry_ecto.ex
bryannaegele: > Also, you can replace `module()` with `Ecto.Repo.t()`

This will compile and publish when it isn't a dependency?

whatyouhide (approved):
Fantastic 💯 I will let @tsloughter or @bryannaegele merge this.

instrumentation/opentelemetry_ecto/lib/opentelemetry_ecto.ex
whatyouhide: Good point. It will compile and publish, but fail Dialyzer. I’m not sure why Ecto is not a dependency of this lib though, but either way yeah we can revert this to be `repo :: module()` for now 🙃

yordis: What should I do then? Add ecto as a deps or revert the change, @bryannaegele?

whatyouhide: @yordis revert the change for now and we can think about Ecto as a dep later on 🙃

bryannaegele: Ecto should continue to not be a dep. You can add a type if `module()` isn't good enough.

whatyouhide: > Ecto should continue to not be a dep

I’m confused as to why but this is not related to the PR really 🙃

bryannaegele: > > Ecto should continue to not be a dep
> 
> I’m confused as to why but this is not related to the PR really 🙃

These libraries in a large org are generally packaged up into shared observability libraries that the org uses as a whole. If you include Ecto as a dep then every project in your org is installing Ecto whether it uses it or not. As such, a general rule for all libraries in contrib we've adopted is "never ship the thing with the instrumentation" wherever possible.

yordis: I cleaned up `_build` and `deps` directory, and even without `ecto` as a deps, it is working thou


Screenshot 2024-08-05 at 2 56 17 PM

@bryannaegele unless I make a non-obvious mistake, I think this is OK to do

bryannaegele: Did you try running dialyzer with only prod deps? It's not set up in ci for this package

yordis: It seems to be working thou

Screenshot 2024-08-05 at 4 47 52 PM

bryannaegele: That has dev dependencies in it. You'd have to set dialyxir to install with prod deps and run it as `MIX_ENV=prod mix dialyzer`.

```
rm -rf _build deps
mix deps.get --only prod
MIX_ENV=prod mix dialyzer
```
You shouldn't see any PLTs for ecto or any of the other dev stuff.

yordis: I did run `rm -rf _build dips` before I ran the test; I do not understand the motivation for using `prod` env here; honestly, I have never done it before.

Regardless, I tried,

Screenshot 2024-08-06 at 11 35 22 AM

As you can see, we need to pull all deps, since `dialyzer` is a `dev` deps, and I had to remove `MIX_ENV=prod` for the same reason.

whatyouhide: If someone will use Dialyzer in prod with this library as a dep and Ecto not being a dep Dialyzer will error out for the unknown type. We don't need to get stuck on this anyway, `module()` is more than fine here and we can push this PR forward 😉

yordis: I replaced the type with `module()` @bryannaegele ready for CR

PR #349, opened 2024-07-13
Fix releases for Xandra and Broadway packages

whatyouhide:


tsloughter (approved):

whatyouhide:
I added a fix for `opentelemetry_broadway`, where the version in `mix.exs` was not defined as `@version` and was screwing up the CI job to release 🙃

PR #356, opened 2024-08-12
HTTP instrumentation header extraction functions

bryannaegele:
Adds several header functions for extracting values for use in semconv attributes and can additionally be used for sampling or other decisions. By default, all info extraction functions follow SemConv's prescribed priority order when the headers are passed as a map since original order cannot be determined.

Users may supply their own sort functions for determining priority. Instrumentation libraries should allow a user to provide their own sort functions. Libaries are encouraged to provide an option for the user to provide a specific list of allowed header names to be used for evaluation. Libraries should filter by that list before calling extract functions if the user provides one.

This does not generate the actual attributes since some values are mixed opt-in with the exception of header attribute values which are 100% opt-in.


tsloughter (approved):
One thing I noticed is the test don't include cases like multiple values int he x-forwarded-for header, so I'd add that to the tests.

bryannaegele:
Will do project rename and hex enhancements in follow-up PR

PR #358, opened 2024-08-13
Renaming http utility lib to otel_http

bryannaegele:
We'd like to rename the unused `opentelemetry_instrumentation_http` library to `otel_http` as we think that namespace has no chance of collision in the API and the utility nature wouldn't pose an issue of moving anything should that ever happen. 

This should make DX nicer with the shorter name.


utilities/otel_http/src/otel_http.erl
```diff
@@ -1,4 +1,4 @@
--module(opentelemetry_instrumentation_http).
+-module(otel_http).
 
 -if(?OTP_RELEASE >= 27).
 -define(MODULEDOC(Str), -moduledoc(Str)).
```
joaothallis, line 4: In [line 42](https://github.com/open-telemetry/opentelemetry-erlang-contrib/pull/358/files#diff-025f7f0c5c5449432a0337f9b9b294a452d9a523cab9e01289fb76ffa56b8229R42) there is a comment with `opentelemetry_instrumentation_http`

tsloughter (approved):

PR #359, opened 2024-08-17
Cowboy SemConv 1.27.0

### LINKED ISSUE #361: Cowboy - SemConv 1.26
Update cowboy to be compliant with SemConv 1.26 which includes numerous breaking changes to observability tooling.

bryannaegele:
SemConv 1.26 compatibility.

This also introduces the equivalent of Go's public endpoint functionality to let users decide when a request should be processed as internal or external.

This introduces a large number of breaking changes from the evolution of the spec since 1.17 and 1.20. Given the stability of the library over time to date and the need to version to a breaking number, we'll set this to 1.0.0 to force users to manually bump the version.

Addresses #183 for cowboy

Resolves #361

Relies on https://github.com/open-telemetry/opentelemetry-erlang/pull/733 and #358


GregMefford (approved):
Looks good overall, just a few small things I noticed.

GregMefford (approved):
I think this is ready to merge.

PR #368, opened 2024-08-20
feat: read db operation name from telemetry_options

yordis:
## Context

Ecto allows to pass `:telemetry_options` as a shared options https://github.com/elixir-ecto/ecto/blob/cba494787b8e98aa0038f6c35e2f44d3661f2620/lib/ecto/repo.ex#L93-L94

I use that to set the `db.operation` OTEL value.

### DIFF
--- a/instrumentation/opentelemetry_ecto/lib/opentelemetry_ecto.ex
+++ b/instrumentation/opentelemetry_ecto/lib/opentelemetry_ecto.ex
@@ -82,12 +82,13 @@ defmodule OpentelemetryEcto do
   def handle_event(
         event,
         measurements,
-        %{query: query, source: source, result: query_result, repo: repo, type: type},
+        %{query: query, source: source, result: query_result, repo: repo, type: type} = meta,
         config
       ) do
     # Doing all this even if the span isn't sampled so the sampler
     # could technically use the attributes to decide if it should sample or not
 
+    options = Map.get(meta, :options, [])
     total_time = measurements.total_time
     end_time = :opentelemetry.timestamp()
     start_time = end_time - total_time
@@ -134,11 +135,15 @@ defmodule OpentelemetryEcto do
 
     db_statement_config = Keyword.get(config, :db_statement, :disabled)
 
+    # TODO: Use OpenTelemetry.SemanticConventions.Trace after 1.26 support
+    db_operation = Keyword.get(options, :"db.operation.name")
+
     attributes =
       base_attributes
       |> add_measurements(measurements, time_unit)
       |> maybe_add_db_statement(db_statement_config, query)
       |> maybe_add_db_system(repo.__adapter__())
+      |> maybe_add_db_operation(db_operation)
       |> add_additional_attributes(additional_attributes)
 
     parent_context =
@@ -199,6 +204,15 @@ defmodule OpentelemetryEcto do
       _, acc ->
         acc
     end)
+  end
+
+  defp maybe_add_db_operation(attributes, nil) do
+    attributes
+  end
+
+  defp maybe_add_db_operation(attributes, db_operation) do
+    # TODO: Use OpenTelemetry.SemanticConventions.Trace after 1.26 support
+    Map.put(attributes, :"db.operation.name", db_operation)
   end
 
   defp maybe_add_db_statement(attributes, :enabled, query) do
--- a/instrumentation/opentelemetry_ecto/mix.exs
+++ b/instrumentation/opentelemetry_ecto/mix.exs
@@ -64,7 +64,8 @@ defmodule OpentelemetryEcto.MixProject do
       {:ecto_sql, ">= 3.0.0", only: [:dev, :test]},
       {:postgrex, ">= 0.15.0", only: [:dev, :test]},
       {:dialyxir, "~> 1.1", only: [:dev, :test], runtime: false},
-      {:opentelemetry_process_propagator, "~> 0.3"}
+      {:opentelemetry_process_propagator, "~> 0.3"},
+      {:opentelemetry_semantic_conventions, "~> 0.2"}
     ]
   end
 end
--- a/instrumentation/opentelemetry_ecto/test/opentelemetry_ecto_test.exs
+++ b/instrumentation/opentelemetry_ecto/test/opentelemetry_ecto_test.exs
@@ -2,6 +2,7 @@ defmodule OpentelemetryEctoTest do
   use ExUnit.Case
   import Ecto.Query
   require OpenTelemetry.Tracer
+  require OpenTelemetry.SemanticConventions.Trace, as: Trace
 
   alias OpentelemetryEcto.TestRepo, as: Repo
   alias OpentelemetryEcto.TestModels.{Comment, User, Post}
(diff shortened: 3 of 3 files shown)


instrumentation/opentelemetry_ecto/lib/opentelemetry_ecto.ex
```diff
@@ -201,6 +205,14 @@ defmodule OpentelemetryEcto do
     end)
   end
 
+  defp maybe_add_db_operation(attributes, nil) do
+    attributes
+  end
+
+  defp maybe_add_db_operation(attributes, db_operation) do
+    Map.put(attributes, Trace.db_operation(), db_operation)
+  end
```
yordis, line 214: In general, should I care about guarding against `nil,` or is that something that downstream would filter out? Or should I even bother sending a `nil` value in case it is nil?

I see `nil` values in my traces, so I want to confirm the general rule around the topic.

instrumentation/opentelemetry_ecto/lib/opentelemetry_ecto.ex
```diff
@@ -201,6 +205,14 @@ defmodule OpentelemetryEcto do
     end)
   end
 
+  defp maybe_add_db_operation(attributes, nil) do
+    attributes
+  end
+
+  defp maybe_add_db_operation(attributes, db_operation) do
+    Map.put(attributes, Trace.db_operation(), db_operation)
```
bryannaegele, line 213: This is not a valid attribute any longer so this would need to be introduced with semconv 1.26. It's also experimental so it'll have to be opt-in.

For `db.operation.name` the value is supposed to be the command or operation being invoked (function name, SQL operation, etc), so I'm not sure it can be changed by the user.

> [5]: It is RECOMMENDED to capture the value as provided by the application without attempting to do any case normalization. If the operation name is parsed from the query text, it SHOULD be the first operation name found in the query. For batch operations, if the individual operations are known to have the same operation name then that operation name SHOULD be used prepended by BATCH , otherwise db.operation.name SHOULD be BATCH or some other database system specific term if more applicable.

bryannaegele (commented):
This attribute is deprecated so we'll need to come back to implementing it after semconv 1.26 is introduced. The new attribute is experimental so it'll have to come in a later pass.

instrumentation/opentelemetry_ecto/lib/opentelemetry_ecto.ex
yordis: Should I hard-code it for now without using `Trace.*`? I intended to use SemConv. I wasn't sure what to do about it.

> For db.operation.name the value is supposed to be the command or operation being invoked (function name, SQL operation, etc), so I'm not sure it can be changed by the user.

I asked for direction in the Slack channel, the wording and examples are confusing me!

instrumentation/opentelemetry_ecto/lib/opentelemetry_ecto.ex
```diff
@@ -82,7 +83,7 @@ defmodule OpentelemetryEcto do
   def handle_event(
         event,
         measurements,
-        %{query: query, source: source, result: query_result, repo: repo, type: type},
+        %{query: query, source: source, result: query_result, repo: repo, type: type, options: options},
```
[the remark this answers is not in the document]
yordis: Based on the docs: https://github.com/elixir-ecto/ecto/blob/cba494787b8e98aa0038f6c35e2f44d3661f2620/lib/ecto/repo.ex#L169-L170 it sounds that it should always be there.

But, it seems to be introduced in `v3.4.0` https://github.com/elixir-ecto/ecto/blob/cba494787b8e98aa0038f6c35e2f44d3661f2620/CHANGELOG.md?plain=1#L606

yordis:
@bryannaegele, given that there is some key under the SemConv (do not get cut up on the one I used), should we wait until the SemConv changes are done, or could we manually add it and add the function later?

yordis:
After reading https://opentelemetry.io/docs/specs/semconv/database/sql/ I am even more confused;

The intent is to add the identity of a given call to the database so that I can identify where the call may be coming from.

- Why can you not use the table name, operation name, or query itself?

The query is dynamic, and it doesn't have a limit of characters, so systems like CloudWatch could not ingest the data as is. Other things like table name and operation name may tell me what happened, but not where it came from, so I can not observe the metric where the optimizations should be performed without hunting for where the query may be coming from.

So, I am not sure what the answer is here.

In the worst case, maybe allow a key to be passed in as "additional attributes."?!

yordis:
From https://opentelemetry.io/docs/specs/semconv/database/database-metrics/

> [5]: It is RECOMMENDED to capture the value as provided by the application without attempting to do any case normalization.

Wouldn't that imply that passing is OK, as I am doing it in the PR?

The second part is about "**If** the operation name is parsed from the query text"

instrumentation/opentelemetry_ecto/lib/opentelemetry_ecto.ex
whatyouhide: @yordis you can look at `opentelemetry_xandra` where I use all hardcoded semcov names exactly because we are waiting for 1.26

bryannaegele: Either way it is an experimental attribute which requires part of the rewrite of accepted options and overall processing. I'm happy to work through what the values can be so that it can go in there but it will not be published in a release before then.

yordis: That works for me!

Based on the description, it could be anything from the app, so I think my intent is correct.

Maybe we should follow up with OTEL folks; regardless, I would love to have a mechanism to have an identity for the operation.

Maybe I should expand the feature to allow passing of any value into the span `telemetry_options: [otel_attributes: [...]] (to make a point)

bryannaegele:
> The intent is to add the identity of a given call to the database so that I can identify where the call may be coming from.

Wouldn't the parent span identify the caller? Trying to understand what you're wanting to achieve just because I believe this field is only for the precise name of the operation being executed rather than the context of the caller. I'm guessing that is the case?
> Other things like table name and operation name may tell me what happened, but not where it came from

> Wouldn't that imply that passing is OK, as I am doing it in the PR?

That line just says we shouldn't modify what is parsed, e.g. `select` shouldn't be modified to `SELECT`

And just so we're on the same page, this library only does spans at this time, so [this section](https://opentelemetry.io/docs/specs/semconv/database/database-spans/) is the relevant one.

Let's see what the WG has to say. A complicating factor here is this value is conditionally part of the span name so any solution has to take that into account for all use cases.

yordis:
@bryannaegele 

Screenshot 2024-08-23 at 12 10 25 PM

yordis:
Related to https://github.com/open-telemetry/semantic-conventions/issues/1365#issuecomment-2387312943

bryannaegele (commented):
This should be addressed with the most recent update

yordis:
@bryannaegele maybe I missed it, but I do not see a way to specify the db operation; sometimes, we do not want to be based on the SQL operation which is what led me here

PR #371, opened 2024-08-30
Instrument batching in opentelemetry_broadway

mashton:
Hooks into Broadway's batch processor start/stop events to add telemetry support to batching. Uses the the `topology` and `batcher` names to construct the span name. Span `kind` is `:internal`.

Unsurprisingly, `OpenTelemetry.SemanticConventions` has no opinions about attribute names for batch processing. Instead we've namespaced a few novel attributes with `broadway.batch` and include message count, failed message count, and successfully message count.

### DIFF
--- a/instrumentation/opentelemetry_broadway/lib/opentelemetry_broadway.ex
+++ b/instrumentation/opentelemetry_broadway/lib/opentelemetry_broadway.ex
@@ -53,6 +53,22 @@ defmodule OpentelemetryBroadway do
         []
       )
 
+    :ok =
+      :telemetry.attach(
+        "#{__MODULE__}.batch_start",
+        [:broadway, :batch_processor, :start],
+        &__MODULE__.handle_batch_start/4,
+        []
+      )
+
+    :ok =
+      :telemetry.attach(
+        "#{__MODULE__}.batch_stop",
+        [:broadway, :batch_processor, :stop],
+        &__MODULE__.handle_batch_stop/4,
+        []
+      )
+
     :ok
   end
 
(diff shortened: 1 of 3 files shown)


mashton:
Hi folks, pardon me if just opening a PR out of the blue is not the done thing here. I didn't see any guidance otherwise. I'd love to contribute to this and other otel projects in the ecosystem.

instrumentation/opentelemetry_broadway/lib/opentelemetry_broadway.ex
```diff
@@ -137,6 +153,58 @@ defmodule OpentelemetryBroadway do
...
+      successful_messages: successful_messages,
+      failed_messages: failed_messages
+    } = metadata,
+    _config
+  ) do
+    status = if length(failed_messages) == 0 do
+      OpenTelemetry.status(:ok)
+    else
+      OpenTelemetry.status(:error, "Batch completed with failed messages")
+    end
```
whatyouhide, line 193: `length/1` always calculates the whole length. Do this instead:

```suggestion
    status = if Enum.empty?(failed_messages) do
      OpenTelemetry.status(:ok)
    else
      OpenTelemetry.status(:error, "Batch completed with failed messages")
    end
```

which will short-circuit instead.

whatyouhide (commented):
This is very old and I’m not really in a place where I can have time to follow up on this, so I'm reviewing just so this falls out of the list of PRs I need to review according to GitHub 😄

yordis:
Superseded by #666 and #675, which landed batch processor instrumentation on `main`: `[:broadway, :batch_processor, :start/:stop]` handlers, `messaging.batch.message_count` plus `messaging.broadway.batch.successful_count` and `messaging.broadway.batch.failed_count`, and an error status when a batch contains failed messages.

I verified the three scenarios from this PR against `main` (successful batch, a message failed inside `handle_batch`, an exception raised inside `handle_batch`) and each one produces the expected batch span with the expected counts and status.

Thanks for the contribution @mashton, and sorry it sat this long. Closing as already covered.

PR #374, opened 2024-09-06
Phoenix Otel semconv 1.27 updates

### LINKED ISSUE #362: Phoenix - SemConv 1.26
Update phoenix to be compliant with SemConv 1.26. Include changes to support bandit as an adapter.

bryannaegele:
Updates SemConv to 1.26 and adds support for Bandit. Using instrumentation for the adapter in use is also no longer optional.

This needs testing particularly around exceptions with the cowboy adapter. The new integration test showed duplicated exception events recorded for bandit and cowboy adapters. This was expected for Bandit but not for Cowboy, so we need some extra testing there. If exceptions are not being recorded properly for Cowboy, we'll add exception handling back for just that adapter.

Resolves #362


bryannaegele:
@mtrudel any idea why bandit http2 integration tests would fail in CI? It works fine locally and it only fails 6/8 of the tests

instrumentation/opentelemetry_phoenix/README.md
```diff
@@ -22,30 +22,18 @@ may be supplied.
 ```elixir
 def deps do
   [
-    {:opentelemetry_phoenix, "~> 1.2"}
+    {:opentelemetry_phoenix, "~> 2.0.0-beta.1"}
```
GregMefford, line 25: I'm not familiar with the release conventions here, but just wanted to confirm that we want to call this beta 1 as opposed to RC 1 or something like that.

instrumentation/opentelemetry_phoenix/lib/opentelemetry_phoenix.ex
```diff
@@ -6,8 +6,9 @@ defmodule OpentelemetryPhoenix do
                       doc: "The endpoint prefix in your endpoint."
                     ],
                     adapter: [
-                      type: {:in, [:cowboy2, :bandit, nil]},
-                      default: nil,
+                      type: {:in, [:cowboy2, :bandit]},
+                      default: :cowboy2,
+                      required: true,
```
GregMefford, line 11: Yeah it's too bad to have a breaking change here, but I agree it makes more sense to force the choice instead of having to also have a default thing built in here.

instrumentation/opentelemetry_phoenix/lib/opentelemetry_phoenix.ex
```diff
@@ -50,12 +53,11 @@ defmodule OpentelemetryPhoenix do
       end
 
   """
-  require OpenTelemetry.Tracer
-  alias OpenTelemetry.SemanticConventions
+  # alias OpenTelemetry.SemConv.Incubating.HTTPAttributes
```
GregMefford, line 56: Is there a reason to leave this here, commented-out?

instrumentation/opentelemetry_phoenix/lib/opentelemetry_phoenix.ex
```diff
@@ -152,186 +132,72 @@ defmodule OpentelemetryPhoenix do
...
   end
 
-  @doc false
-  def handle_endpoint_start(_event, _measurements, meta, config) do
-    Process.put({:otel_phoenix, :adapter}, config.adapter)
-
-    case adapter() do
-      :cowboy2 ->
-        cowboy2_start()
+  # TODO: do we still need exception handling? Only when cowboy?
```
GregMefford, line 135: Don't we still want to have the `handle_router_dispatch_exception`?

instrumentation/opentelemetry_phoenix/mix.exs
```diff
@@ -62,18 +62,33 @@ defmodule OpentelemetryPhoenix.MixProject do
...
+      {:opentelemetry_semantic_conventions, "~> 1.27"},
+      {:otel_http, "~> 0.2"},
+      {:telemetry, "~> 1.0"},
       {:plug, ">= 1.11.0"},
       {:cowboy_telemetry, "~> 0.4", only: [:dev, :test]},
-      {:opentelemetry, "~> 1.0", only: [:dev, :test]},
-      {:opentelemetry_exporter, "~> 1.0", only: [:dev, :test]},
+      {:opentelemetry_exporter, "~> 1.7", only: [:dev, :test]},
+      {:opentelemetry, "~> 1.4", only: [:dev, :test]},
+      {:opentelemetry_bandit,
```
GregMefford, line 76: Should we get these git dependencies resolved before merging?

GregMefford (approved):
Nothing major, but some questions to iron out.

instrumentation/opentelemetry_phoenix/README.md
bryannaegele: We can do rc.

instrumentation/opentelemetry_phoenix/mix.exs
bryannaegele: Yeah. It has to wait until those are published. Can't test or publish if the packages aren't on hex.

instrumentation/opentelemetry_phoenix/lib/opentelemetry_phoenix.ex
bryannaegele: Right. With bandit support now it just doesn't make sense to not force it and this is a breaking release anyhow, so time to do it.

instrumentation/opentelemetry_phoenix/lib/opentelemetry_phoenix.ex
bryannaegele: Did not appear so in testing.

instrumentation/opentelemetry_phoenix/lib/opentelemetry_phoenix.ex
```diff
@@ -152,186 +131,72 @@ defmodule OpentelemetryPhoenix do
...
-    # end the Phoenix span
-    OpentelemetryTelemetry.end_telemetry_span(@tracer_id, meta)
-  end
-
   @doc false
   def handle_router_dispatch_start(_event, _measurements, meta, _config) do
     attributes = %{
       :"phoenix.plug" => meta.plug,
       :"phoenix.action" => meta.plug_opts,
-      SemanticConventions.Trace.http_route() => meta.route
```
sc-yan, line 247: @bryannaegele  hi just wonder if there is any particular reason that we stop passing `http.route` to attribute? because from the [spec](https://github.com/open-telemetry/opentelemetry-specification/pull/3165), if there is no `http.route`, the span name would be only `{http.method}`.

danschultzer: HTTP route is something Bandit or Cowboy should set which is the URI, the `meta.route` is the route template from Phoenix.

sc-yan: @danschultzer hi thanks so much for replying. could you take a look at this issue please? coz as far as I checked, it seems Bandit didn't set `http.route` somehow.

danschultzer: Oh sorry I was thinking about something completely different (`http.path`). You are right, this shouldn't have been removed.

OTel semantics v1.27 specifies that either `http.route` or `http.template` can be used: https://github.com/open-telemetry/semantic-conventions/blob/v1.27.0/docs/http/http-spans.md#name

> The <span id="target-placeholder">`{target}`</span> SHOULD be one of the following:
> 
> - [`http.route`](/docs/attributes-registry/http.md) for HTTP Server spans
> - [`url.template`](/docs/attributes-registry/url.md) for HTTP Client spans if enabled and available (Experimental)
> - Other value MAY be provided through custom hooks at span start time or later.

`http.route` should be used as this is a server span not a client span.

sc-yan: thanks so much for confirming! I'm so sorry that I forgot to paste the URL of my github issue. here it is.
https://github.com/open-telemetry/opentelemetry-erlang-contrib/issues/464
I also created a [pull request](https://github.com/open-telemetry/opentelemetry-erlang-contrib/pull/465) to fix it.

could someone check the issue and PR please? or who should I mention?

bryannaegele: Is your `/healthz` endpoint templated or static? Is it defined in the Phoenix router or in a plug before the router is hit?

sc-yan: @bryannaegele  `healthz` endpoint is static. it's defined in `router.ex` like this
 ` get "/healthz", XxxxxWeb.HealthzController, :index`

PR #375, opened 2024-09-13
Req 1.27 SemConv

### LINKED ISSUE #351: OpentelemetryReq: log when path_params not set and continue
Continuing from a discussion I had on Slack with @bryannaegele.

This pertains to the behaviour present in the OpentelemetryReq library, where the request can be halted under the following circumstances:
- `:no_path_params` option is unset
- `path_params` has not been used in the `Req.Request`

The main issue from my point of view is that the instrumentation is changing application behaviour by halting the request [here](https://github.com/open-telemetry/opentelemetry-erlang-contrib/blob/main/instrumentation/opentelemetry_req/lib/opentelemetry_req.ex#L216-L222).

**Describe the solution you'd like**
- remove the `:no_path_params` option in favour of the branch `no_path_params: true`.
- In other words, fall back from path_params `HTTP GET /foo/:bar` to simply `HTTP GET`
- Log (Bryan suggested to `info`, I'd suggest `debug`) when this fallback occurs.
- _Never_ halt a request.

**Describe alternatives you've considered**
None

**Additional context**
None

### LINKED ISSUE #363: Req - SemConv 1.26
Update req to be compliant with SemConv 1.26

bryannaegele:
Updates Req to SemConv 1.26. This creates a number of breaking changes in order to comply with the new SemConv.

Given previous stability, breaking changes, and otel http semconv stability, this will mark a 1.0 release, as well.

Addresses #183 for Req

Resolves #351 
Resolves #363


tsloughter:
If it wasn't 1.0 then it wasn't stable. No need to do a major version bump if it was 0.x.

instrumentation/opentelemetry_req/lib/opentelemetry_req.ex
```diff
@@ -30,39 +108,82 @@ defmodule OpentelemetryReq do
...
+  >
+  """
+
+  @spec attach(Req.Request.t(), options()) :: Req.Request.t()
   def attach(%Req.Request{} = request, options \\ []) do
+    opts = NimbleOptions.validate!(options, @options_schema) |> Enum.into(%{})
+
+    config =
+      opts
+      |> Enum.into(%{})
```
GregMefford, line 162: We don't need to duplicate the `Enum.into(%{})` in the config processing, but also maybe we could just combine it all into one pipe? It looks like nothing is using `opts` except the `Enum.member?` which seems like it could just as easily check for the value being in the `config` Map.

```suggestion
    config =
      options
      |> NimbleOptions.validate!(@options_schema)
      |> Map.new()
```

instrumentation/opentelemetry_req/lib/opentelemetry_req.ex
```diff
@@ -30,39 +108,82 @@ defmodule OpentelemetryReq do
...
+        attrs
+        |> Enum.filter(&elem(&1, 1))
+        |> Keyword.keys()
+      end)
+      |> then(fn config ->
+        if Enum.member?(opts.opt_in_attrs, {URLAttributes.url_template(), true}) do
+          Map.put(config, :url_template_enabled, true)
+        else
+          Map.put(config, :url_template_enabled, false)
+        end
```
GregMefford, line 173: What's the deal with renaming this one in particular? ([cross-reference to SemConv](https://github.com/open-telemetry/opentelemetry-erlang/blob/main/apps/opentelemetry_semantic_conventions/lib/incubating/attributes/url_attributes.ex#L265))

instrumentation/opentelemetry_req/lib/opentelemetry_req.ex
```diff
@@ -86,14 +207,31 @@ defmodule OpentelemetryReq do
...
-    Tracer.set_attributes(attrs)
+    opt_in =
+      %{
+        HTTPAttributes.http_response_body_size() => extract_response_body_size(response)
+      }
+      |> Map.take(config.opt_in_attrs)
 
     if response.status >= 400 do
-      OpenTelemetry.Tracer.set_status(OpenTelemetry.status(:error, ""))
+      Tracer.set_status(OpenTelemetry.status(:error, ""))
```
GregMefford, line 219: Looks like it was already like this, but IMO it shouldn't be an error span necessarily just because the server successfully returned a 404 for example, if that's what the caller expected. In that case, the client did its job correctly, so there was no error. In fact, the server that the client called probably did as well unless it was a 5xx error, so it shouldn't be an error on its span either. Does OTel say we should do it this way for HTTP clients, or is this just something we've chosen to do here?

instrumentation/opentelemetry_req/lib/opentelemetry_req.ex
```diff
@@ -139,98 +291,162 @@ defmodule OpentelemetryReq do
...
 
-  defp maybe_append_req_content_length(attrs, req) do
-    case Req.Request.get_header(req, "content-length") do
+  defp extract_port(%{port: port}) when is_integer(port), do: port
+
+  defp extract_port(%{scheme: scheme}) do
+    case scheme do
+      nil -> 80
+      "http" -> 80
+      "https" -> 8080
```
GregMefford, line 330:
```suggestion
      "https" -> 443
```

instrumentation/opentelemetry_req/lib/opentelemetry_req.ex
```diff
@@ -139,98 +291,162 @@ defmodule OpentelemetryReq do
...
+  defp extract_user_agent(request) do
+    case Req.Request.get_header(request, "user-agent") do
       [] ->
-        attrs
+        ""
 
-      [length] ->
-        Map.put(attrs, Trace.http_request_content_length(), length)
+      [user_agent] ->
+        user_agent
```
GregMefford, line 354: It would be weird, but maybe we should not blow up if there are multiple UA headers, but just use the first one?
```suggestion
      [user_agent | _] ->
        user_agent
```

instrumentation/opentelemetry_req/lib/opentelemetry_req.ex
```diff
@@ -139,98 +291,162 @@ defmodule OpentelemetryReq do
...
-        attrs
+        0
+
+      [length_str] when is_binary(length_str) ->
+        # Req sets this as a string but should be an integer
+        # https://www.rfc-editor.org/rfc/rfc9110.html#name-content-length
+        # https://opentelemetry.io/docs/specs/semconv/attributes-registry/http
+        String.to_integer(length_str)
+
+      [length] ->
```
GregMefford, line 369: Similar to the above, and I imagine it would never happen in practice, but what if there were multiple headers returned? Makes me think maybe what we really want here is a `get_first_header` API where it returns the first one if there are one or more, or `nil` if it's not there, so make the call sites less likely to one day crash due to weird data.

instrumentation/opentelemetry_req/test/opentelemetry_req_test.exs
```diff
@@ -17,48 +25,365 @@ defmodule OpentelemetryReqTest do
...
+  end
+
+  describe "span name" do
+    test "with no path params" do
+      Req.get!(client(),
+        plug: &ok_resp/1,
+        url: "/"
+      )
+
+      assert_receive {:span, span(name: :GET)}
```
GregMefford, line 235: Is it weird that the name is an atom here and a binary when the path params are parsed? Feels surprising to me.

GregMefford (commented):
I think this is mostly ready to go, thanks so much for the work here! ❤️ 
I just had a few comments to think about / tweak before I'm ready to give it a ✅

instrumentation/opentelemetry_req/lib/opentelemetry_req.ex
bryannaegele: Not sure I follow. Renaming of what?

instrumentation/opentelemetry_req/lib/opentelemetry_req.ex
bryannaegele: It's an error for HTTP clients.

> For HTTP status codes in the 4xx range span status MUST be left unset in case of SpanKind.SERVER
and SHOULD be set to Error in case of SpanKind.CLIENT.

https://hexdocs.pm/opentelemetry_semantic_conventions/http-spans.html#status

instrumentation/opentelemetry_req/test/opentelemetry_req_test.exs
bryannaegele: We don't `to_atom` anything for users as a general rule. I'm not sure I'd want to deviate from it since sobelow would probably gripe but I don't know that for a fact. What do you think?

GregMefford: Right, I don't think we'd convert to an atom, but I was kind of expecting that the span name would always be a string.

instrumentation/opentelemetry_req/lib/opentelemetry_req.ex
GregMefford: It looks like SemConv defines this field as `:"url.template"`, but this code is pulling that value out and putting it in the `:url_template_enabled` key in `config`. I think I just misread this as renaming it in the attributes themselves, but it's actually just setting a key in the configuration based on that attribute name, so it's fine! 👍

instrumentation/opentelemetry_req/lib/opentelemetry_req.ex
GregMefford: IMO that's a bummer and not something I would want to have in my traces personally, so I guess that's why it's a `SHOULD` instead of a `MUST`. But as a library, I guess we have to decide whether we think it's right to follow that convention, since it's not required. My vote would be to not set it as an error unless it was like a timeout or connection error or something where there's no response at all. If it's a 5xx, I would be fine with it being marked as an error in the client, just to call attention to it even though it's not Req's fault if the server returns a 504 or something, but I think it's totally normal for 404s to happen all the time when you're just querying an API for "does this thing exist?" and it just says "no" as a 4040 - you don't need to call special attention to that.

bryannaegele: Users are allowed to mark a span ok so they can hook into the request and mark it should they choose to. We could look at adding a response step to let folks supply a function but this PR isn't seeking to change any existing functionality. I think that's a fine feature request.

instrumentation/opentelemetry_req/lib/opentelemetry_req.ex
bryannaegele: It's just an internal config key to not have to parse the attrs on every request, essentially a precalculation.

instrumentation/opentelemetry_req/lib/opentelemetry_req.ex
```diff
@@ -1,26 +1,104 @@
...
+  ## Req Path Params
+
+  It is strongly encouraged to use `req_path_params` module which should be registered before this step.
+  This allows the span name to include the `{target}` portion of the span name described in the
+  [HTTP Span Name guidelines](https://opentelemetry.io/docs/specs/semconv/http/http-spans/#name).
+
+  > #### Requirements {: .info}
+  >
+  > * `req_path_params` module must be registered before `OpentelemetryReq` step
+  > * `URLAttributes.url_template()` opt-in attribute must be set to `true`
```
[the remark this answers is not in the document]
bryannaegele: Good catch. This was outdated as its a built-in step now and doesn't require any special config any longer.

instrumentation/opentelemetry_req/lib/opentelemetry_req.ex
GregMefford: Yep, that makes sense! ✅

instrumentation/opentelemetry_req/test/opentelemetry_req_test.exs
GregMefford: Similar to the other comment, I don't think this is a blocker if that's already how it was working before, but I agree that it would be something worth considering as an option in a future PR where we can focus on just that aspect.

GregMefford (approved):
I think this one is good to merge.

PR #398, opened 2024-10-18
🐛 Fix handler on streams

rparcus:
Fixes https://github.com/open-telemetry/opentelemetry-erlang-contrib/issues/327

### DIFF
--- a/instrumentation/opentelemetry_finch/lib/opentelemetry_finch.ex
+++ b/instrumentation/opentelemetry_finch/lib/opentelemetry_finch.ex
@@ -42,10 +42,25 @@ defmodule OpentelemetryFinch do
     end_time = :opentelemetry.timestamp()
     start_time = end_time - duration
 
+    # status =
+    #   case meta.result do
+    #     {:ok, response} -> response.status
+    #     _ -> 0
+    #   end
+
     status =
       case meta.result do
-        {:ok, response} -> response.status
-        _ -> 0
+        {:ok, response} when is_map(response) ->
+          response.status
+
+        {:ok, {_request, response}} when is_map(response) ->
+          response.status
+
+        {:ok, {status, _headers, _body}} when is_integer(status) ->
+          status
+
+        _ ->
+          0
       end
 
     url = build_url(meta.request.scheme, meta.request.host, meta.request.port, meta.request.path)
--- a/instrumentation/opentelemetry_finch/test/opentelemetry_finch_test.exs
+++ b/instrumentation/opentelemetry_finch/test/opentelemetry_finch_test.exs
@@ -51,6 +51,110 @@ defmodule OpentelemetryFinchTest do
            } = :otel_attributes.map(attributes)
   end
 
+  test "records span on streamed requests where the acc is a 3 elements tuple", %{
+    bypass: bypass,
+    test: finch_name
+  } do
+    Bypass.expect_once(bypass, "GET", "/", fn conn ->
+      Plug.Conn.send_resp(conn, 200, "OK")
+    end)
+
+    OpentelemetryFinch.setup()
+
+    _conn = start_supervised!({Finch, name: finch_name})
+
+    acc = {nil, [], ""}
+
+    fun = fn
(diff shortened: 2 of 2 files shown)


rparcus:
Hello 👋
Without this fix it is impossible to instrument anything that uses streaming + finch.
Is there anything I can do to further help with the review process?

instrumentation/opentelemetry_finch/lib/opentelemetry_finch.ex
```diff
@@ -44,8 +44,17 @@ defmodule OpentelemetryFinch do
 
     status =
       case meta.result do
-        {:ok, response} -> response.status
-        _ -> 0
+        {:ok, response} when is_map(response) ->
```
rparcus, line 47: The main issue was here. Not every response contains a status key as it depends on the implementation of the streaming callback functions.

A fallback case is also added. In this fix, I'm not trying to come up with a general solution but just one that works for the patterns I saw. The tests use callbacks that return values in the patterns I was already working with.

lud-wj: I guess you could just match on the presence of the status:

```elixir
case meta.result do
  {:ok, %{status: status}} when integer(status) -> status
  {:ok, {_request, %{status: status}}} when integer(status) -> status
  {:ok, {status, _headers, _body}} when is_integer(status) -> status
  _ -> 0
end
```

Because there is a risk that even when `response` is a map, `response.status` may not be set. As far as I can tell that part of the event is the stream accumulator and it could be anything, any term.

lud-wj:
+1 for this PR.

We are using the Langchain lib with streams, and it is not compatible with OTEL.

rparcus:
Thanks for the suggestion above. I think it's great @lud-wj.
Done in 9ac5b88

> +1 for this PR.
> 
> We are using the Langchain lib with streams, and it is not compatible with OTEL.

That is exactly why this PR exists ✨

instrumentation/opentelemetry_finch/lib/opentelemetry_finch.ex
lud-wj: (edit fixed my clauses :') )

bwan-nan (approved):

gmile:
Would be nice to have this merged 🙏

maxdrift:
@bryannaegele, it seems you've been reviewing and merging the `opentelemetry_finch` PRs in this repo. Any chance you could review this one too? 🙏 
Thank you in advance!

instrumentation/opentelemetry_finch/test/opentelemetry_finch_test.exs
```diff
@@ -51,6 +51,110 @@ defmodule OpentelemetryFinchTest do
...
+  } do
+    Bypass.expect_once(bypass, "GET", "/", fn conn ->
+      Plug.Conn.send_resp(conn, 200, "OK")
+    end)
+
+    OpentelemetryFinch.setup()
+
+    _conn = start_supervised!({Finch, name: finch_name})
+
+    acc = {nil, %{body: "", headers: %{}, trailers: [], status: 200}}
```
yordis, line 106:
```suggestion
    acc = {nil, %{body: "", headers: %{}, trailers: %{}, status: 200}}
```

is trailer an actual map?

rparcus: image

And then in Mint we have:

```elixir
@typedoc """
  HTTP headers.

  Headers are sent and received as lists of two-element tuples containing two strings,
  the header name and header value.
  """
  @type headers() :: [{header_name :: String.t(), header_value :: String.t()}]
  ```
  
  They are basically maps, but they are initialized as empty lists everywhere in Finch so I'd rather keep it `[]`

rparcus: Maybe I missed something that is more relevant to actual runtime?

PR #424, opened 2024-12-05
fix: add support for dataloader kv sources

btkostner:
This fixes the dataloader source breaking using KV sources instead of Ecto. Right now these tests currently break because the code assumes we are running with an Ecto source.


yordis (approved):

yordis:
Moved to #762

PR #426, opened 2024-12-10
OTel Resource Detector for AWS

### LINKED ISSUE #101: Resource Detector: AWS

danschultzer:
Adds OTel Resource Detector for AWS, with support for ECS.

It's essentially a mirror of the [JS version](https://github.com/open-telemetry/opentelemetry-js-contrib/blob/main/detectors/node/opentelemetry-resource-detector-aws/src/detectors/AwsEcsDetectorSync.ts).

Resolves #101

### DIFF
--- a/.github/hex-packages.json
+++ b/.github/hex-packages.json
@@ -152,6 +152,15 @@
     "language": "elixir",
     "authorizedUsers": ["bryannaegele", "tsloughter"]
   },
+  "resource_detector_aws": {
+    "workingDirectory": "detectors/opentelemetry_resource_detector_aws",
+    "name": "Resource Detector for AWS",
+    "packageName": "opentelemetry_resource_detector_aws",
+    "tagPrefix": "opentelemetry-resource-detector-aws-v",
+    "buildTool": "rebar3",
+    "language": "erlang",
+    "authorizedUsers": ["danschultzer"]
+  },
   "tesla": {
     "workingDirectory": "instrumentation/opentelemetry_tesla",
     "name": "Tesla Instrumentation",
--- a/detectors/opentelemetry_resource_detector_aws/src/otel_resource_aws_ecs.erl
+++ b/detectors/opentelemetry_resource_detector_aws/src/otel_resource_aws_ecs.erl
@@ -1,0 +1,140 @@
+-module(otel_resource_aws_ecs).
+
+-behavior(otel_resource_detector).
+
+-export([get_resource/1]).
+
+-include_lib("kernel/include/logger.hrl").
+-include_lib("opentelemetry_semantic_conventions/include/incubating/attributes/cloud_attributes.hrl").
+-include_lib("opentelemetry_semantic_conventions/include/incubating/attributes/container_attributes.hrl").
+-include_lib("opentelemetry_semantic_conventions/include/incubating/attributes/aws_attributes.hrl").
+
+-define(OS_ENV, "ECS_CONTAINER_METADATA_URI_V4").
+-define(DEFAULT_CGROUP_PATH, "/proc/self/cgroup").
+
+%% @private
+-spec get_resource(list()) -> otel_resource:t() | no_return().
+get_resource(Config) ->
+    otel_resource:create(fetch(Config)).
+
+fetch(Config) ->
+    case os:getenv(?OS_ENV) of
+        false ->
+          erlang:error('Not running in ECS environment, ECS_CONTAINER_METADATA_URI_V4 not set.');
+        _MetadataV4URL ->
+          [
+            {?CLOUD_PROVIDER, atom_to_list(?CLOUD_PROVIDER_VALUES_AWS)},
+            {?CLOUD_PLATFORM, atom_to_list(?CLOUD_PLATFORM_VALUES_AWS_ECS)} |
+            aws_ecs_attributes(Config)
+          ]
+    end.
+
+aws_ecs_attributes(Config) ->
+  {ok, Hostname} = inet:gethostname(),
+
+  ContainerIDAttributes =
+    case fetch_container_id_attributes(Config) of
+      {ok, ContainerIDAttributesResult} ->
+        ContainerIDAttributesResult;
+      {error, ContainerIDAttributesError} ->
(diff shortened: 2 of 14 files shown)


detectors/opentelemetry_resource_detector_aws/src/otel_resource_aws_ecs.erl
```diff
@@ -0,0 +1,139 @@
+-module(otel_resource_aws_ecs).
```
danschultzer, line 1: The naming convention here follows the naming convention in opentelemetry with [`otel_resource_app_env`](https://github.com/open-telemetry/opentelemetry-erlang/blob/main/apps/opentelemetry/src/otel_resource_env_var.erl) and [`otel_resource_env_var`](https://github.com/open-telemetry/opentelemetry-erlang/blob/main/apps/opentelemetry/src/otel_resource_env_var.erl)

.github/hex-packages.json
```diff
@@ -152,6 +152,15 @@
...
     "authorizedUsers": ["bryannaegele", "tsloughter"]
   },
+  "resource_detector_aws": {
+    "workingDirectory": "detectors/opentelemetry_resource_detector_aws",
+    "name": "Resource Detector for AWS",
+    "packageName": "opentelemetry_resource_detector_aws",
+    "tagPrefix": "opentelemetry-resource-detector-aws-v",
+    "buildTool": "rebar3",
+    "language": "erlang",
+    "authorizedUsers": ["danschultzer"]
```
danschultzer, line 162: I assume I have to manage the releases for this, but lmk if either of you @bryannaegele @tsloughter should be added here.

danschultzer:
@bryannaegele @tsloughter can I get a review of this PR?

danschultzer:
I'll maintain it separately, released here:

https://github.com/danschultzer/opentelemetry-resource-detector-aws
https://hex.pm/packages/opentelemetry_resource_detector_aws

tsloughter:
@danschultzer sorry we never reviewed this :(. Can you please reopen. This should really be in contrib. Even if you feel you need to publish it third party for now I'd like this PR to remain open so we can review it and get it merged eventually and released with the rest. This will help to ensure as semantic conventions change they are kept in sync.

detectors/opentelemetry_resource_detector_aws/src/otel_resource_aws_ecs.erl
```diff
@@ -0,0 +1,140 @@
...
+
+fetch_container_id_attributes(Config) ->
+  Path = proplists:get_value(cgroup_path, Config, ?DEFAULT_CGROUP_PATH),
+  
+  case file:read_file(Path) of
+      {ok, Data} ->
+          Lines = binary:split(Data, <<"\n">>, [global]),
+          ContainerId = lists:foldl(fun
+            (Line, none) ->
+              leength,
```
tsloughter, line 63: What is going on here?

detectors/opentelemetry_resource_detector_aws/src/otel_resource_aws_ecs.erl
```diff
@@ -0,0 +1,140 @@
...
+  
+  [{?CONTAINER_NAME, Hostname}] ++ ContainerIDAttributes ++ MetadataV4Attributes.
+
+fetch_container_id_attributes(Config) ->
+  Path = proplists:get_value(cgroup_path, Config, ?DEFAULT_CGROUP_PATH),
+  
+  case file:read_file(Path) of
+      {ok, Data} ->
+          Lines = binary:split(Data, <<"\n">>, [global]),
+          ContainerId = lists:foldl(fun
```
tsloughter, line 61: Can you add a comment for how this foldl is supposed to work? It seems to loop through Lines from a file until one is found that is greater than 64 bytes and then take the last 64 bytes? It then continues to iterate through the list but doing nothing since its already found the thing it wanted? You could use `lists:search` and then extract the part of the Line you want, if that is the case.

Why is that the line with the container id?

danschultzer: It's a pattern laid down in other instrumentation SDKs with go and js:

https://github.com/open-telemetry/opentelemetry-go-contrib/blob/main/detectors/aws/ecs/ecs.go#L250-L254
https://github.com/open-telemetry/opentelemetry-js-contrib/blob/main/detectors/node/opentelemetry-resource-detector-aws/src/detectors/AwsEcsDetector.ts#L157-L177

I don't know exactly why it's done this way. We just want to find the container id which is a 64 byte hex binary, but personally I would have pattern matched on `/docker` instead, but I don't know how cgroup file may look. So I've just kept it like the others.

detectors/opentelemetry_resource_detector_aws/src/otel_resource_aws_ecs.erl
danschultzer: Good catch, not sure how that assigns got there.

danschultzer:
> @danschultzer sorry we never reviewed this :(. Can you please reopen. This should really be in contrib. Even if you feel you need to publish it third party for now I'd like this PR to remain open so we can review it and get it merged eventually and released with the rest. This will help to ensure as semantic conventions change they are kept in sync.

Thanks but I rather wait until PRs here have quicker turnaround time than months. I'm spending time keeping my branches here clean and up to date so they can be easily reviewed and merged, and I don't want to have to maintain two versions of the same thing for any significant amount of time. It'll diverge and be tougher and tougher keeping parity. When there is enough attention here I will push what I got so we can port it over.

I understand that it's not easy to maintain OSS (I have several projects myself I wish I could spend more time on), but I'm hitting my head against hard deadlines now. I've pushed it as long as a could to give enough time for review and getting this in but I'm out of time now.

I'll be closing my other PR's and maintaining a separate fork to unblock work. I will push upstream changes in the future when this moves quicker and I have more time (depending how divergent the fork becomes).

tsloughter:
Ok. Curious, do you have forks of just contrib apps or core ones as well?

We definitely need help with contrib. This is the hardest, by far, project I'm a part of to maintain because it is not just reviewing the code but cross checking with every detail in the semantic conventions that they must follow.

PR #430, opened 2024-12-14
Update OpenTelemetryEcto attributes to conform with semantic conventions 1.27

### LINKED ISSUE #364: Ecto -SemConv 1.26
Update ecto to be compliant with SemConv 1.26

danschultzer:
Stacked on #429 which needs to get in first. Check the latest commit for the changes here.

This conforms the attributes and span names to the 1.27 docs: https://github.com/open-telemetry/semantic-conventions/blob/v1.27.0/docs/database/database-spans.md

It does introduce a bunch of breaking changes so let me know if backwards compability is required. I can keep the old attributes, but not sure if we should just go with major release instead since there's a lot of things that are different now?

I can also update the changelog in this PR.

## Changes

- `db.name`/`db.instance` -> `db.namespace`
- `source` -> `db.collection.name`
- `db.statement` -> `db.query.text`
- `db.url` -> `server.address`/ `server.port`
- `total_time_` -> `db.client.operation.duration` (in seconds as float)
- `idle_time_` -> `db.client.connection.create_time` (in seconds as float)
- `queue_time_` -> `db.client.connection.wait_time` (in seconds as float)
- `query_time_` -> `db.client.connection.use_time` (in seconds as float)
- The `: db_statement` config option to `:db_query` (not necessary but it conforms closer to the otel attribute name)
- Span name changed from `ecto.repo.query` to `{db.operation.name} {target}` or just `{target}` if it is a general db operation conforming to 1.27 guidelines

## Removes
- `db.type`
- `decode_time_` (I couldn't find a metric that matched it)
- The `:span_prefix` config option
- The `:time_unit` config option

## Adds
- `db.operation.name`

Resolves #364

### DIFF
--- a/instrumentation/opentelemetry_ecto/lib/opentelemetry_ecto.ex
+++ b/instrumentation/opentelemetry_ecto/lib/opentelemetry_ecto.ex
@@ -20,17 +20,16 @@ defmodule OpentelemetryEcto do
 
   require OpenTelemetry.Tracer
 
-  alias OpenTelemetry.SemConv.Incubating.DBAttributes
+  alias OpenTelemetry.SemConv.Incubating.{DBAttributes, Metrics.DBMetrics}
+  alias OpenTelemetry.SemConv.{ErrorAttributes, ServerAttributes}
 
   @typedoc """
   Option that you can pass to `setup/2`.
   """
   @typedoc since: "1.3.0"
   @type setup_option() ::
-          {:time_unit, System.time_unit()}
-          | {:span_prefix, String.t()}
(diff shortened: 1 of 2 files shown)


bryannaegele:
@danschultzer I'm going to work on your fork to move this along quickly so we don't have to bounce it back and forth for weeks.

danschultzer:
@bryannaegele yup, go for it.

danschultzer:
@bryannaegele FYI I've rebased this on main, formatted, and fixed test to get this succeed in CI (some of the deps updated broke tests).

PR #431, opened 2024-12-14
Add LiveView URL template attribute

danschultzer:
This PR consists of two commits, one that refactors the LiveView test to use something that's closer to how LiveView itself tests sockets, and a second that parses and adds the URL template attribute to match the attribute being added by the router dispatch event.

I can split it into two PRs if that's preferred.

## Adds

- `url.template` to the liveview spans


yordis:
Superseded by #809 and #810, which are merged. Thanks for the original work here.

PR #435, opened 2024-12-15
Update OpentelemetryOban to use semantic conventions 1.27

danschultzer:
We have to override the `:semantic_conversions` dep because oban requires `~> 0.2` but most otel libraries here uses 1.27.


bryannaegele:
@danschultzer can you please confirm you have thoroughly reviewed the [Messaging Specification](https://hexdocs.pm/opentelemetry_semantic_conventions/messaging.html) for any new or changed attributes that are applicable as well as any other sections? As far as I can tell this was not performed since nothing has changed in the setup for opting into experimental attirbutes which is essentially the entire messaging specification. Please refer to bandit, phoenix, cowboy, etc to familiarize yourself with the expected API.

It is not sufficient to just bump the dependency.

Thanks

danschultzer:
Yes, it's all in the followup PR #436, which adheres as much as possible to 1.27 specs for messaging.

But it'll introduce many breaking change.

This PR only focuses on updating the dependency, but all the attributes are backwards compatible so I don't think it's necessary to force the changes here?

FWIW the dependency was not locked to 0.2, but allowed the range `>= 0.2 and < 1.0`, and most of the current libraries allows `>= 1.27 and < 2.0`.

danschultzer:
Also one thing on all my PRs, I can update the changelog if it's preferred to do in the PR's. I'm not sure how the releases are managed.

bryannaegele:
Right, 1.27 does introduce a lot of breaking changes which is why we want it done properly and not just publishing a version with an updated dep with no actual updates. By going to 1.27 we are saying it is 100% 1.27 compliant to our knowledge. That's why this is a difficult update and folks have to exercise some patience. It's the most difficult update we'll have to make and hopefully the last for a long time.

danschultzer:
I see, but then do we need to fixate the semantic conventions version to `1.27`? Because it's at `1.29` now with more breaking changes, so whenever the semconv library gets updated we end up in the same place with drift if we continue using the `~>` version requirement.

I was thinking of using this PR in a minor release to bump the semconv version (we could make the version here be `~> 0.2 or ~> 1.27` to really make it backwards compatible), and a major release where all the breaking changes are included to conform it to 1.27 guidelines.

It gets the current dependency conflict out of the way without also breaking everything for the end-user (but they will get deprecation warnings which I think is ok).

danschultzer:
@bryannaegele 1.18 now causes deprecation warnings to show up so the compile with warnings as errors fails in CI. We could close this PR and move on to #435 where all the attributes will be changed as well, or hardcode the attribute keys here to get this non breaking change in first.

danschultzer:
I pushed a commit to hardcode the keys to have this go green on CI, but you can work on this fork as you see fit @bryannaegele.

luigi-discoup:
Are there any updates on this? I'm stuck on an old version of `opentelemetry_cowboy` and `opentelemetry_phoenix` due to `opentelemetry_oban` requiring `opentelemetry_semantic_conventions` `v0.2`.  

Is there anything I can do to help close this PR?  

Thank you.

grzuy:
For the record, this other pull request seems to be similar https://github.com/open-telemetry/opentelemetry-erlang-contrib/pull/528.

PR #437, opened 2024-12-15
Trace all Oban telemtry events

danschultzer:
Stacked on #435 which needs to get in first. Check the latest commit for the changes here.

Oban triggers a lot of the ecto otel spans. I think we should set up spans for all internal events so we can have oban parent spans for these ecto spans.

The span names is just the telemetry event joined with `.`, e.g. `oban.peer.election`. Maybe there is a better convention for span names for these internal processes, or there may be some these telemetry events that should have a specific span names (and attributes).

I've also changed the telemetry handler id, I usually see tuples being used rather than binary.

Sidenote: I think a good good future refactor would be to combine all the handlers into just one module instead, I don't think it is necessary to have this concept of `Handler.attach` vs just calling `attach_many` with all the oban events we want to listen to.


instrumentation/opentelemetry_oban/lib/opentelemetry_oban/internal_handler.ex
```diff
@@ -0,0 +1,64 @@
...
+      []
+    )
+  end
+
+  def handle_oban_event(event, _measurements, metadata, _config) do
+    [op | rest] = Enum.reverse(event)
+
+    case op do
+      :start ->
+        OpentelemetryTelemetry.start_telemetry_span(__MODULE__, Enum.join(Enum.reverse(rest), "."), metadata, %{kind: :consumer})
```
danschultzer, line 50: Probably should add `messing.system` here to make it easy to filter?

danschultzer: Same goes for the plugin handler.

PR #438, opened 2024-12-16
[WIP] Add route to the span name to Phoenix socket requests

danschultzer:
This is WIP as it needs an upstream fix in Phoenix: https://github.com/phoenixframework/phoenix/pull/6019

This will add a route template to the attributes, and use that for the span name. We may want to add code attributes from #432 though I'm not sure as we deal with both the `socket` module that is the transport (e.g. `Phoenix.Transports.WebSocket`) and the `user socket` module that's the actual socket passed into the `socket` macro.


instrumentation/opentelemetry_phoenix/lib/opentelemetry_phoenix.ex
```diff
@@ -145,6 +156,18 @@ defmodule OpentelemetryPhoenix do
     |> OpenTelemetry.Ctx.attach()
   end
 
+  @doc false
+  def handle_socket_dispatch_start(_event, _measurements, meta, config) do
+    if config.adapter == :cowboy2, do: cowboy2_start()
```
danschultzer, line 161: This was necessary as it didn't update the actual span. I assume the difference from the router event is that the socket handling happens much earlier than router, and there is some handling of cowboy process in phoenix that links it correctly.

yordis:
Phoenix doesn't emit `[:phoenix, :socket_dispatch, :start]`, so this can't land as is. Closing for now, happy to revisit if that event goes upstream. Thanks for the work here.

Issue #302, opened 2024-03-17
Lost parent context when Phoenix LiveView assign_async is used


arathunku:
**Describe the bug**

When Phoenix.LiveView `assign_async/3` is used, any spans from async task will not have parent context, **even when [OpentelemetryProcessPropagator](https://hexdocs.pm/opentelemetry_process_propagator/OpentelemetryProcessPropagator.html)** is used. This means that any Ecto query is detached span.

**Expected behavior**

I'd expect parent context to be propagated. 

**Additional context**

Repro script:

```elixir
#!/usr/bin/env elixir
Application.put_env(:sample, Example.Endpoint,
  http: [ip: {127, 0, 0, 1}, port: 5001],
  server: true,
  live_view: [signing_salt: "you-are-the-best"],
  secret_key_base: String.duplicate("a", 64)
)

Application.put_env(:opentelemetry, :traces_exporter, {:otel_exporter_stdout, []})

Mix.install([
  {:plug_cowboy, "~> 2.5"},
  {:jason, "~> 1.0"},
  {:phoenix, "~> 1.7"},
  {:phoenix_live_view, "~> 0.20.1"},
  {:opentelemetry, "~> 1.3"},
  {:opentelemetry_api, "~> 1.2"},
  {:opentelemetry_exporter, "~> 1.6"},
  {:opentelemetry_process_propagator, "~> 0.3.0"},
  {:opentelemetry_phoenix, "~> 1.1"}
])

defmodule Example.ErrorView do
  def render(template, _), do: Phoenix.Controller.status_message_from_template(template)
end

defmodule Example.HomeLive do
  use Phoenix.LiveView, layout: {__MODULE__, :live}
  require OpenTelemetry.Tracer, as: Tracer

  def mount(_params, _session, socket) do
    IO.inspect({Tracer.current_span_ctx(), connected?(socket)})

    span_ctx = Tracer.current_span_ctx()

    {:ok,
     socket
     |> assign_async(:items, fn ->
       # detached spans are expected here I guess
       Tracer.with_span "span-without-propagation" do
         {:ok, %{items: []}}
       end
     end)
     |> assign_async(:special_items, fn ->
       with_telemetry_ctx(fn ->
         # BUG! this is also detached span?
         Tracer.with_span "span-with-propagation" do
           {:ok, %{special_items: []}}
         end
       end)
     end)
     |> assign_async(:special_items_ctx, fn ->
[...]

PR #439, opened 2024-12-16
Add LiveView async wrapper functions for process propagation in OpentelemetryPhoenix

danschultzer:
This adds helper macros to add propagation to the LiveView async macros.  It follows the same in-place replacement as ProcessPropagator and Oban, so you must require `OpentelemetryPhoenix.LiveView` in the liveview and replace.

Resolves: https://github.com/open-telemetry/opentelemetry-erlang-contrib/issues/302


danschultzer:
Turning this to draft until I can figure out how to make this not depend on live_view as per https://github.com/open-telemetry/opentelemetry-erlang-contrib/pull/433#issuecomment-2547269703:

> Our instrumentation libraries cannot include dependencies on the library being instrumented. There are a few reasons for this but as you come up with ideas or proposals just know that is a hard constraint.

Maybe this should be a propagator library instead? Fits better with the the process propagator (in that case oban will need to be the same).

PR #440, opened 2024-12-19
add_opentelementry_sqlcommenter_package

dkuku:
As discussed - the sqlcommenter package.


yordis:
@dkuku is the org be able to publish the pkg? who owns the Hex pkg permission right now?

yordis:
@dkuku also, we need the CLA to be signed

dkuku:
@yordis I'm the owner - we can change the ownership. I spoke about this with @tsloughter

yordis:
@dkuku if you figure out Hex, and CI, I can help to merge things in 😃

dkuku:
package [transferred ](https://hex.pm/packages/opentelemetry_sqlcommenter) to opentelemetry

yordis:
A few things before this can go in:

mix.exs deps are all unscoped, so they land in every consumer's tree:

- Drop postgrex entirely. It isn't referenced anywhere in lib/, and ~> 0.19.3 pulls in CVE-2026-32687 (HIGH, SQL injection in Postgrex.Notifications.listen/3) plus decimal CVE-2026-32686 transitively.
- Add {:opentelemetry_api, "~> 1.4"}. It's the actual runtime dependency, since OpenTelemetry.Tracer lives there, but right now it only resolves transitively.
- Move opentelemetry and opentelemetry_exporter to only: [:dev, :test], and ex_doc to only: [:dev], runtime: false.

Package metadata should match the other instrumentations (see instrumentation/opentelemetry_redix/mix.exs):

- links.GitHub still points at github.com/dkuku/opentelemetry_sqlcommenter; it should be the monorepo path. Same for adding source_url and a docs block.
- Add package[:files], a LICENSE, and a CHANGELOG.md; none are in the directory today.
- Bump the version. 0.1.1 is already published on Hex, so release-drafter can't cut a tag at that version.

Minor: the @moduledoc duplicates itself, with the README pasted in and then a second "Adds OpenTelemetry trace context to SQL queries as comments" section repeating it. encode_flags/1's integer fallback uses Integer.to_string(x, 16), which emits uppercase hex where W3C traceparent wants lowercase (only reachable for flag values other than 0/1). And [comment: ..., prepare: :unnamed] ++ opts silently shadows a caller-supplied prepare:.

dkuku:
I addresed all the comments. 
> And [comment: ..., prepare: :unnamed] ++ opts silently shadows a caller-supplied prepare:.

for this one I added a comment in the readme - having a comment with a named statement does not make sense because the every query will be different so no point of caching it.

instrumentation/opentelemetry_sqlcommenter/LICENSE
```diff
@@ -0,0 +1,202 @@
...
+      To apply the Apache License to your work, attach the following
+      boilerplate notice, with the fields enclosed by brackets "[]"
+      replaced with your own identifying information. (Don't include
+      the brackets!)  The text should be enclosed in the appropriate
+      comment syntax for the file format. We also recommend that a
+      file or class name and description of purpose be included on the
+      same "printed page" as the copyright notice for easier
+      identification within third-party archives.
+
+   Copyright [yyyy] [name of copyright owner]
```
yordis: what should the value be for the brackets?

dkuku: replaced with the license file from opentelemetry_bandit

yordis (approved):

PR #445, opened 2024-12-28
Ignore .DS_Store files

danschultzer:
Developing on MacOS we have `.DS_Store` files generated for for all directories opened in Finder.


bryannaegele:
Hey Dan. You can set this in your global gitignore.

https://gist.github.com/subfuzion/db7f57fff2fb6998a16c

danschultzer:
Ah right that makes sense.

PR #448, opened 2025-01-08
Fix OpentelemetryPhoenix tests

danschultzer:
Resolves the current failing tests in main. There seems to be some race condition introduced.

Instead of starting and stopping the telemetry app it's easier to just detach all listeners when the test ends.


yordis:
Closing as superseded: the race-condition fix (detaching telemetry handlers in on_exit instead of stopping the :telemetry app) landed on main via 6e9570f, and the full opentelemetry_phoenix suite passes on current main (37 tests, 0 failures). The remaining diff is stylistic only. Thanks @danschultzer!

PR #449, opened 2025-01-09
Refactor integration test

danschultzer:
Stacked on #448.

Updates the dependencies being tested with, uses Phoenix formatter rules, and refactors the integration test with the following changes:

1. No need to set up two ports at the same time since we only test a single endpoint for each test
2. Use test tags for opts/config
3. Detach only the telemetry handlers we know are attached
4. DRY setup and request handling


yordis:
Thanks for this, @danschultzer, and sorry it sat so long. The integration tests have been reworked a few times since, so this no longer applies. Closing it out.

PR #462, opened 2025-01-27
OpenTelemetry Bandit: Add a nil check for status

JohnDoneth:
It appears that WebSockets are reported as having a `nil` conn.status to the telemetry handler. Due to the conditional in the telemetry handler `nil` statuses are interpreted as errors (when they are functioning as expected) due to `nil >= 500` evaluating to true.


bryannaegele:
Can you sign the cla?

yordis:
Fixed by https://github.com/mtrudel/bandit/pull/466

PR #469, opened 2025-02-19
Conform attributes and span name to v1.27 semantics for OpentelemetryTesla

ShapovalovaIrina:
Conform attributes and span name to v1.27 semantics for OpentelemetryTesla.

Follows the guidelines set out in https://github.com/open-telemetry/semantic-conventions/blob/v1.27.0/docs/http/http-spans.md.

## Changed

- Span name `{custom hook}` changed to `{method} {target/custom hook}`
- Span name `HTTP {method}` changed to `{method}`
- Span name `{path}` changed to `{method} {target/template}`
- Set `error.type` attribute when span status is error
- Rename attribute `http.method` -> `http.request.method`
- Rename attribute `http.url` -> `url.full`
- Rename attribute `http.status_code` -> `http.response.status_code`
- Rename attribute `net.host.name` -> `server.address`
- Rename attribute `http.response.content.length` -> `http.response.body.size`

## Added

- Add attribute `server.port`

## Removed

- Remove attribute `http.target`
- Remove attribute `http.scheme`

## Questions

1. According to the guideline `error.type` should be set to exception type or component-specific error identifier if response was not received. Also error.type value should be predictable and should have low cardinality. Tesla uses adapters for different HTTP clients, so some errors are quite difficult to standardize. So for errors like this, I've set the value to `undefined` for now, иut I'm not sure if this is a good decision.

2. `url.full`, `http.request.method`, `server.address`, `server.port` attributes can theoretically be set before the request is made. In this case, these parameters will be specified even if the request failed with `{:error, any()}`.
In practice, the `url.full` value may not be fully known until the request is completed, since there may be such middlewares as `Tesla.Middleware.PathParams` or `Tesla.Middleware.Query`.
But in the `opentelememtry_tesla 2.4.0` these attributes are set after request is made, so perhaps I shouldn't worry about this behaviour.


bryannaegele:
Can you please refer to the updates to Req for all the appropriate API changes that need to be implemented here? 

Same things would apply to Finch, as well. 

Thanks!

ShapovalovaIrina:
> Can you please refer to the updates to Req for all the appropriate API changes that need to be implemented here?

Just to clarify, are you talking about changes in https://github.com/open-telemetry/opentelemetry-erlang-contrib/pull/375?

ShapovalovaIrina:
@bryannaegele made changes to Tesla and Finch referring to Req updates. Will be waiting for feedback :)

ShapovalovaIrina:
@ricardoccpaiva Hello! Sorry to bother you, but as far as I understand you are the codeowner of the opentelemetry_tesla, so could you please leave some feedback on the PR?

instrumentation/opentelemetry_tesla/lib/middleware/opentelemetry_tesla_middleware.ex
```diff
@@ -17,40 +17,51 @@ defmodule Tesla.Middleware.OpenTelemetry do
...
-    span_name_fun.(env)
+    "#{http_method(env.method)} #{span_name_fun.(env)}"
   end
 
   defp get_span_name(env, _) do
     case env.opts[:path_params] do
-      nil -> "HTTP #{http_method(env.method)}"
-      _ -> URI.parse(env.url).path
+      nil -> "#{http_method(env.method)}"
+      _ -> "#{http_method(env.method)} #{URI.parse(env.url).path}"
```
tsloughter, line 67: If this really should be `.path` can you put a comment on why it is here? It is something to do with `path_params` not being empty? Does it being not nil mean `.path` is a template of a path and not a high cardinality path?

kmc-firework: `.path` does not include query parameters so it shouldn't be high cardinality if that's what you're wondering.

example:
```elixir
iex(1)> URI.parse("https://google.com/things?find=true")
%URI{
  scheme: "https",
  authority: "google.com",
  userinfo: nil,
  host: "google.com",
  port: 443,
  path: "/things",
  query: "find=true",
  fragment: nil
}
```

tsloughter: Path is kept out kept out in the semantic conventions because it is potentially high cardinality. Which is why the conventions say to only include it as a template/route if available, like, `GET /things/:id`.

tsloughter: What is `path_params`?

ShapovalovaIrina: `path_params` is an option for the Tesla client to substitute parameters into the path template. [Tesla.Middleware.PathParams](https://hexdocs.pm/tesla/Tesla.Middleware.PathParams.html) is used for this.

For this piece of code, the logic is as follows: if `path_params` is nil, then the path is not presented as a template and potentially has a high cardinality, so only the method is used for the span name. If `path_params` is present, then the path is presented as a template and can be substituted into the span name.

I will leave a comment in the code for this logic.

ShapovalovaIrina: Examples of span names from tests:

```elixir
client =
  Tesla.client([
    {Tesla.Middleware.BaseUrl, base_url},
    Tesla.Middleware.OpenTelemetry,
    Tesla.Middleware.PathParams
  ])

Tesla.get(client, "/users/:id", opts: [path_params: [id: "3"]])

assert_receive {:span, span(name: "GET /users/:id", attributes: _attributes)}
```

```elixir
client =
  Tesla.client([
    {Tesla.Middleware.BaseUrl, base_url},
    Tesla.Middleware.OpenTelemetry
  ])

Tesla.get(client, "/users/")

assert_receive {:span, span(name: "GET", attributes: _attributes)}
```

Since we don't analyze what's in the path (template or full path) and rely only on path_params, the second option uses a generic span name "GET" to avoid potentially high cardinality.

hoyon:
Sorry to nudge but is there any update on this PR? The current version of opentelemetry_tesla depends on opentelemetry_semantic_conventions `~> 0.2` which makes it difficult to use with other otel libraries which already depend on `~> 1.27` without resorting to using an override.

tsloughter:
@hoyon no worries. I think we can probably get this one merged in soon. I need to double check some things. Hopefully tomorrow unless its decided some options need to change or be added around the attributes included.

tsloughter (approved):

ShapovalovaIrina:
Thanks for merging PR :) 

Do you have any updates about a new release with these changes?

tsloughter:
Release done.

PR #470, opened 2025-02-20
Conform attributes and span name to v1.27 semantics for OpentelemetryFinch

ShapovalovaIrina:
Conform attributes and span name to v1.27 semantics for OpentelemetryFinch.

Follows the guidelines set out in https://github.com/open-telemetry/semantic-conventions/blob/v1.27.0/docs/http/http-spans.md.

## Changed

- Change span name `HTTP {method}` -> `{method}`
- Set span status to error, when HTTP status code is in 4xx or 5xx
- Set `error.type` attribute when span status is error
- Rename attribute `http.method` -> `http.request.method`
- Rename attribute `http.url` -> `url.full`
- Rename attribute `http.status_code` -> `http.response.status_code`
- Rename attribute `net.host.name` -> `server.address`
- Rename attribute `http.scheme` -> `url.scheme`

## Added

- Add attribute `server.port`

## Removed

- Remove attribute `http.target`


ShapovalovaIrina:
@bryannaegele Hello! Sorry for bothering you, I couldn't find any information about who is opentelemetry_finch codeowner, so I'm contacting you. Can you please tag the right person or leave some feedback on PR?

bryannaegele:
Hey @ShapovalovaIrina , this is great!

Two requests:
* list out the aliases as fully qualified (makes it easier to find references during updates)
* use the semantic conventions for matches in tests so we don't have to know the values of the keys - [example](https://github.com/open-telemetry/opentelemetry-erlang-contrib/blob/main/instrumentation/opentelemetry_phoenix/test/integration_test.exs#L232-L244)

ShapovalovaIrina:
> Two requests:
> 
> * list out the aliases as fully qualified (makes it easier to find references during updates)
> * use the semantic conventions for matches in tests so we don't have to know the values of the keys - [example](https://github.com/open-telemetry/opentelemetry-erlang-contrib/blob/main/instrumentation/opentelemetry_phoenix/test/integration_test.exs#L232-L244)

@bryannaegele made the changes that were requested :)

Similar changes have also been made for OpentelemetryTesla PR https://github.com/open-telemetry/opentelemetry-erlang-contrib/pull/469.

ShapovalovaIrina:
Hi @bryannaegele, do you have an estimate for when you might be able to take a look at this? No rush, just wanted to check 🙂  Thanks!

instrumentation/opentelemetry_finch/lib/opentelemetry_finch.ex
```diff
@@ -41,45 +144,198 @@ defmodule OpentelemetryFinch do
...
+    set_span_status(s, meta.result, status)
+
+    OpenTelemetry.Span.end_span(s)
+  end
+
+  defp get_otel_config(request_private) do
+    request_private
+    |> Map.get(:otel, %{})
+    |> NimbleOptions.validate!(@options_schema)
+  end
```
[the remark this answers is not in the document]
yordis: This is a good point, not sure what we would like to happen here

instrumentation/opentelemetry_finch/lib/opentelemetry_finch.ex
```diff
@@ -41,45 +144,211 @@ defmodule OpentelemetryFinch do
...
-      Trace.http_method() => meta.request.method,
-      Trace.http_status_code() => status
-    }
+    otel_config = get_otel_config(meta.request.private)
+    span_name = span_name(meta.request, otel_config)
+    status = extract_status(meta.result)
 
     s =
-      OpenTelemetry.Tracer.start_span("HTTP #{meta.request.method}", %{
+      Tracer.start_span(span_name, %{
```
yordis, line 153: Should we use `OpentelemetryTelemetry` instead?

tsloughter: What is it needed for?

instrumentation/opentelemetry_finch/lib/opentelemetry_finch.ex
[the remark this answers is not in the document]
tsloughter: It should log but not skip span creation in my opinion.

instrumentation/opentelemetry_finch/lib/opentelemetry_finch.ex
yordis: Doesn't that handle the span creation and closing more idiomatic? Not sure, I always use it when the events come from `:telemetry` directly

tsloughter: I forget. I was thinking it was if the parent could be in another process, but that is something else actually, right?

tsloughter: This actually ends the span in the same function? So I don't think anything extra is needed.

yordis:
@tsloughter merge?

ShapovalovaIrina:
> Can we get the conflicts resolved? 

Done ✅

yordis (approved):

Issue #508, opened 2025-05-12
OpentelemetryOba.JobHandler - Trace - Parent Span being ignored


linqueta:
**Describe the bug**
I have a piece of code where I have an OpenTelemetry Trace Span already in context, and I set them in the field `meta` to the following Oban Jobs. Once the Oban Jobs starts to process, it triggers `OpentelemetryOban.JobHandler.handle_job_start/4`, but it's ignoring a possible parent span, ignoring the three entirely (`OpenTelemetry.Tracer.set_current_span(:undefined)`) 

A simple change to `OpenTelemetry.Tracer.set_current_span(parent)` makes it work, having a parent or not.

**Expected behavior**
Respect the parent span passed into `job_meta`, calling to `OpenTelemetry.Tracer.set_current_span(parent)` 

**Additional context**

tsloughter:
Is it setting it to a Link? The active span that creates the job should be linked to and is not considered a parent. Though I'd have to verify if its not in the semconv that it could be optionally set as parent and not a Link.

PR #509, opened 2025-05-12
fix: Avoid ignoring Parent Span

linqueta:
As discussed on #508, `Oban.JobHandler` seems not to be setting the right span, ignoring the parent, causing a break in the traces for a flow. This fix allows the parent span to be included in the traces. 

In case a parent span doesn't exist, it'll have the value `:undefined`, exactly what has been set by default in the current code.

### DIFF
--- a/instrumentation/opentelemetry_oban/lib/opentelemetry_oban/job_handler.ex
+++ b/instrumentation/opentelemetry_oban/lib/opentelemetry_oban/job_handler.ex
@@ -57,7 +57,7 @@ defmodule OpentelemetryOban.JobHandler do
     :otel_propagator_text_map.extract(Map.to_list(job_meta))
     parent = OpenTelemetry.Tracer.current_span_ctx()
     links = if parent == :undefined, do: [], else: [OpenTelemetry.link(parent)]
-    OpenTelemetry.Tracer.set_current_span(:undefined)
+    OpenTelemetry.Tracer.set_current_span(parent)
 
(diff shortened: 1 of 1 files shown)


instrumentation/opentelemetry_oban/lib/opentelemetry_oban/job_handler.ex
```diff
@@ -57,7 +57,7 @@ defmodule OpentelemetryOban.JobHandler do
     :otel_propagator_text_map.extract(Map.to_list(job_meta))
     parent = OpenTelemetry.Tracer.current_span_ctx()
     links = if parent == :undefined, do: [], else: [OpenTelemetry.link(parent)]
-    OpenTelemetry.Tracer.set_current_span(:undefined)
+    OpenTelemetry.Tracer.set_current_span(parent)
```
indrekj, line 60: IIRC, I did it because we already create an Opentelemetry link. If we also set the parent, then we have both a link and a direct relationship. Can you link from the opentelemetry docs that this should not be a link?

linqueta: I understand @indrekj , I'm not sure if it shouldn't be a link or not, I just evaluated how it was logged at Datadog, and I missed the parent in the whole tree of traces.
So, I don't have a solid argument to say if it should or not. In case not we can close the PR.

PR #511, opened 2025-05-23
add attempted_at, workflow_id and queue to span attributes

xantrac:
Add 
- attempted_at
- queue
- workflow_id (oban pro)
to the span attributes


yordis:
Rebased onto current main to resolve conflicts from the messaging semconv and span_relationship changes that landed after this was opened. The three attributes (queue, attempted_at, workflow_id) are unchanged; tests pass.

yordis (approved):

PR #515, opened 2025-06-03
Use public interfaces in Bandit module

mtrudel:
Previously, the Bandit support module used non-public interfaces within Bandit to obtain the HTTP protocol version and peer address/port info. These interfaces were factored away in Bandit 1.7, since they were never part of the declared public API. 

This PR updates the Bandit support module to use publicly accessible methods for obtaining this info. 

I've not done any extensive in-situ testing of this, though the suite is green for both Bandit 1.6 and 1.7 versions.


bryannaegele:
I cut it.

You just go to the actions and run the hex release workflow for the package. If it needs a version update then it will open a PR for you.

PR #436, opened 2024-12-15
Conform attributes to v1.27 semantics for OpentelemetryOban

danschultzer:
Stacked on #435 which needs to get in first. Check the latest commit for the changes here.

Follows the guideslines set out in https://github.com/open-telemetry/semantic-conventions/blob/v1.27.0/docs/messaging/messaging-spans.md#destinations

This does introduce a bunch of breaking changes so let me know if backwards compability is required. I can keep the old attributes, but not sure if we should just go with major release instead since there's a lot of things that are different now? Same question as in #430

## Changes

- `messaging.destination` -> `messaging.consumer.group.name`
- `messaging.destination.kind` is removed
- Span names `{destination} {operation}` changed to`{operation} {destination}`
- Span names `{plugin} process` changed to `oban.plugin {plugin}`
- Span names `Oban bulk insert` changed to `{operation}` or `{operation} {destination}` depending whether all are the same job type

## Added

- `messaging.destination.name` (the worker module)
- `messaging.operation.name`
- `messaging.operation.type`
- `messaging.message.id`

There's also  more to this to get it to be compliant with 1.27 conventions. A larger one is that we probably need to deal with creation context? https://github.com/open-telemetry/semantic-conventions/blob/v1.27.0/docs/messaging/messaging-spans.md#context-propagation

---

Related #367

PR #528, opened 2025-07-11
fix(opentelemetry_oban): Update semantic conventions in opentelemetry_oban

mhanberg:
### What does this PR do?

Updates the opentelemetry_semantic_conventions dependency of opentelemetry_oban to 1.27.

~I wasn't sure exactly what the replacement is for the "messaging_destination_kind" key, so I deleted it. Please correct me if this was the wrong decision.~ messaging destination kind was indeed deleted, as they didn't find a use case for it.

- Adds the operation type to job insertion as well as job handling
- includes message id, which is the job id
- Migrates some fields that were renamed.

### DIFF
--- a/instrumentation/opentelemetry_oban/lib/opentelemetry_oban.ex
+++ b/instrumentation/opentelemetry_oban/lib/opentelemetry_oban.ex
@@ -103,7 +103,7 @@ defmodule OpentelemetryOban do
     # changesets in insert_all can include different workers and different
     # queues. This means we cannot provide much information here, but we can
     # still record the insert and propagate the context information.
-    OpenTelemetry.Tracer.with_span :"Oban bulk insert", kind: :producer do
+    OpenTelemetry.Tracer.with_span "send", kind: :producer do
       changesets = Enum.map(changesets, &add_tracing_information_to_meta/1)
       Oban.insert_all(name, changesets)
     end
--- a/instrumentation/opentelemetry_oban/test/opentelemetry_oban_test.exs
+++ b/instrumentation/opentelemetry_oban/test/opentelemetry_oban_test.exs
@@ -354,7 +354,7 @@ defmodule OpentelemetryObanTest do
 
     assert_receive {:span,
                     span(
-                      name: :"Oban bulk insert",
+                      name: "send",
                       attributes: _attributes,
                       trace_id: send_trace_id,
                       span_id: send_span_id,


instrumentation/opentelemetry_oban/test/opentelemetry_oban/plugin_handler_test.exs
```diff
@@ -103,7 +103,7 @@ defmodule OpentelemetryOban.PluginHandlerTest do
 
     [
       event(
-        name: "exception",
+        name: :exception,
```
grzuy, line 151: This was already fixed separately in https://github.com/open-telemetry/opentelemetry-erlang-contrib/pull/558

instrumentation/opentelemetry_oban/test/opentelemetry_oban_test.exs
```diff
@@ -175,22 +178,23 @@ defmodule OpentelemetryObanTest do
...
+             "messaging.operation.type": :process,
+             "messaging.system": :oban,
+             "messaging.message.id": ^id,
+             "messaging.client.id": "TestJobThatReturnsError"
            } = :otel_attributes.map(attributes)
 
     [
       event(
-        name: "exception",
+        name: :exception,
```
grzuy, line 200: This was already fixed separately in https://github.com/open-telemetry/opentelemetry-erlang-contrib/pull/558

instrumentation/opentelemetry_oban/test/opentelemetry_oban_test.exs
```diff
@@ -324,7 +329,7 @@ defmodule OpentelemetryObanTest do
...
```
grzuy, line 336: This was already fixed separately in https://github.com/open-telemetry/opentelemetry-erlang-contrib/pull/558

instrumentation/opentelemetry_oban/lib/opentelemetry_oban/job_handler.ex
grzuy: Seems like we also need to update the span name per https://opentelemetry.io/docs/specs/semconv/messaging/messaging-spans/#span-name?

instrumentation/opentelemetry_oban/lib/opentelemetry_oban/job_handler.ex
```diff
@@ -60,10 +58,12 @@ defmodule OpentelemetryOban.JobHandler do
...
-      Trace.messaging_system() => :oban,
-      Trace.messaging_destination() => queue,
-      Trace.messaging_destination_kind() => :queue,
-      Trace.messaging_operation() => :process,
+      MessagingAttributes.messaging_system() => :oban,
+      MessagingAttributes.messaging_message_id() => id,
+      MessagingAttributes.messaging_client_id() => worker,
+      MessagingAttributes.messaging_destination_name() => queue,
+      MessagingAttributes.messaging_operation_type() =>
+        MessagingAttributes.messaging_operation_type_values().process,
```
grzuy, line 67: Does the warning message displayed in the top of https://opentelemetry.io/docs/specs/semconv/messaging/ means that the migration needs to provide a `OTEL_SEMCONV_STABILITY_OPT_IN` env var that allows for previous or newer "messaging" conventions?

Image

instrumentation/opentelemetry_oban/lib/opentelemetry_oban.ex
```diff
@@ -130,15 +129,18 @@ defmodule OpentelemetryOban do
     worker = Changeset.get_field(changeset, :worker)
 
     %{
-      Trace.messaging_system() => :oban,
-      Trace.messaging_destination() => queue,
-      Trace.messaging_destination_kind() => :queue,
+      MessagingAttributes.messaging_system() => :oban,
+      MessagingAttributes.messaging_client_id() => worker,
+      MessagingAttributes.messaging_destination_name() => queue,
```
grzuy, line 135: I see here that you're keeping the existing interpretation that a "Destination" is an Oban queue.

For what is worth and for the record, noting that this other pull request (https://github.com/open-telemetry/opentelemetry-erlang-contrib/pull/436#discussion_r2499351968) is making a different interpretation in that a "Destination" is the Oban Worker and the "Consumer group" is the Oban queue.

PR #529, opened 2025-07-11
feat(phoenix): trace LiveView `render`

grzuy:
Hi there :wave: 

It seems these events are not traced because Phoenix LiveView started emitting these events (https://github.com/phoenixframework/phoenix_live_view/pull/2907) a few months after the liveview integration was done here in OpentelemetryPhoenix (https://github.com/open-telemetry/opentelemetry-erlang-contrib/pull/198).

Anyway, adding, so they are traced and anything downstream the render has the proper parent span linking.


yordis:
Superseded by #622, which was merged and covers the same `[:phoenix, :live_view, :render, ...]` telemetry events (plus `live_component.update`). Closing, thanks for raising this!

PR #540, opened 2025-08-14
feat:(opentelemetry_broadway): add context propagation integration

yordis:


tsloughter:
Nice :+1: I'll review this today.

instrumentation/opentelemetry_broadway/lib/opentelemetry_broadway.ex
```diff
@@ -66,11 +113,13 @@ defmodule OpentelemetryBroadway do
           name: name,
           message: %Broadway.Message{} = message
         } = metadata,
-        _config
+        config
       ) do
     span_name = "#{inspect(topology_name)}/#{Atom.to_string(processor_key)} process"
     client_id = inspect(name)
 
+    {_parent_ctx, links} = get_propagated_ctx(message, config)
```
tsloughter: Why not just not return a 2-tuple and only return the list of links?

instrumentation/opentelemetry_broadway/README.md
```diff
@@ -18,6 +20,39 @@ def start(_type, _args) do
 end
 ```
 
+### With Trace Propagation
```
tsloughter, line 23: Are there backends that don't support metadata? If possible we should list out which ones it doesn't work with and let them know they'll have to package up the context into their messages in that case. Though maybe all of them do support metadata? That'd be nice.

instrumentation/opentelemetry_broadway/lib/opentelemetry_broadway.ex
yordis: 🤦🏻 I forgot, I used to clear up the current span, let me fix that!

instrumentation/opentelemetry_broadway/README.md
yordis: Hey @tsloughter I added most of the documetnation to the module documentation itself,

Is that OK?

Honestly, for a long time I stopped adding so much to the README file other than basic installations, but I rather put the docs in the modules.

I added example of the producers that you may need to update (rabbitmq, sqs) since other like kafka already put the metadata in the right place.

We can evolve the docs over time

instrumentation/opentelemetry_broadway/README.md
```diff
@@ -18,6 +20,19 @@ def start(_type, _args) do
...
+
+For Broadway pipelines that need distributed tracing context extraction from message headers:
```
tsloughter: I think it should mention that it creates links, rather than just saying it extracts context.

yordis:
Hey @tsloughter any updates here?

yordis:
@whatyouhide could you share some blessing here? 🙏🏻

tsloughter:
I think this should be enabled by default, https://github.com/open-telemetry/semantic-conventions/blob/main/docs/messaging/messaging-spans.md

Another piece, which can wait for a future PR, is it should optionally be able to make the producer the parent and not just a link.

yordis:
@whatyouhide you are owner, it seems you must approve it

tsloughter (approved):

instrumentation/opentelemetry_broadway/lib/opentelemetry_broadway.ex
```diff
@@ -14,8 +16,58 @@ defmodule OpentelemetryBroadway do
...
+
+  For Broadway pipelines that need distributed tracing context extraction from message headers/attributes:
+
+      def start(_type, _args) do
+        :ok = OpentelemetryBroadway.setup(propagation: true)
+
+        # ...
+      end
+
+  **Important**: When using trace propagation, your producer must be configured to extract headers/attributes.
```
whatyouhide:
```suggestion
  > #### Extracting Headers and Attributes {: .info}
  When using trace propagation, your producer must be configured to extract headers/attributes. See below for instructions.
```

instrumentation/opentelemetry_broadway/lib/opentelemetry_broadway.ex
```diff
@@ -139,4 +213,53 @@ defmodule OpentelemetryBroadway do
...
+
+    # Create links to parent if it exists
+    case parent_ctx do
+      :undefined -> []
+      _ -> [OpenTelemetry.link(parent_ctx)]
+    end
+  end
+
+  # RabbitMQ: headers are in metadata.headers as a list
+  defp get_message_headers(%Broadway.Message{metadata: %{headers: headers}}) when is_list(headers), do: headers
```
whatyouhide, line 246: Should we think about providing a way for extending this (via another option for example) to avoid coupling strictly to these two adapters *and* allowing more adapters if necessary? I’m thinking of the Redis adapter for example.

`:propagation` could be:

```elixir
propagation: :broadway_rabbitmq | :broadway_sqs | (Broadway.Message.t() -> [{String.t(), String.t()}])
```

or something.

yordis: @whatyouhide the tricky bit with using an atom is that this is handling all producers (intentionally), so you could have any combination of things.

I do like the idea of the function callback thou. At the same time, if covering the 90% of the case by a simply pattern matching does the work. I rather do that, and if people ask, we can do a non-breaking change for them to take control.

If you feel strong about it, I do not have strong opinions, just prefer that things work out of the box if the code required is that minimal.

whatyouhide: Ah yeah multiple producers of course, I was not thinking about that. Okay, that sounds good, let's go out of the box with support for RabbitMQ and SQS and evolve through non-breaking changes if we need to.

whatyouhide (approved):

Issue #551, opened 2025-09-10
Remove references to Collector's batch processor in opentelemetry-erlang-contrib


mx-psi:
There are four references that could be removed: https://github.com/search?q=repo%3Aopen-telemetry%2Fopentelemetry-erlang-contrib%20%2Fbatch%3A%2F&type=code

The idea is to replace as follows:

<table>
  <tr>
<td>

**Before:**

```yaml
receivers:
  foo:

processors:
  batch:

exporters:
  foo:

service:
  pipelines:
    metrics:
      receivers: [foo]
      processors: [batch]
      exporters: [foo]
```

</td>

<td>

**After:**

```yaml
receivers:
  foo:



exporters:
  foo:
    sending_queue:
      batch:

service:
  pipelines:
    metrics:
      receivers: [foo]
      exporters: [foo]

```

</td>
  </tr>
</table>

tsloughter:
@mx-psi should they be replaced with `sending_queue` on the exporter, or not?

mx-psi:
@tsloughter Yeah, specially if there are specific options on the batch processor configuration

PR #552, opened 2025-09-16
Remove batch processor references from OpenTelemetry Collector config

Frapschen:
issue: https://github.com/open-telemetry/opentelemetry-erlang-contrib/issues/551


mx-psi (commented):
Would be good to add

```
sending_queue:
  batch:
```

on each exporter as I did in the example in #551 (sorry, I did not put that in at first)

mx-psi (approved):

tsloughter (approved):

PR #554, opened 2025-10-01
feat(opentelemetry_grpc): add initial setup for opentelemetry_grpc

yordis:
Signed-off-by: Yordis Prieto <yordis.prieto@gmail.com>


instrumentation/opentelemetry_grpc/README.md
```diff
@@ -0,0 +1,141 @@
+# OpentelemetryGrpc
+
+[![EEF Observability WG project](https://img.shields.io/badge/EEF-Observability-black)](https://github.com/erlef/eef-observability-wg)
+[![Hex.pm](https://img.shields.io/hexpm/v/opentelemetry_grpc)](https://hex.pm/packages/opentelemetry_grpc)
+
+OpenTelemetry instrumentation for gRPC clients and servers in Elixir.
```
sleipnir:
```suggestion
`OpenTelemetry` instrumentation for gRPC clients and servers in Elixir, enabling distributed tracing, error tracking, and context propagation across services."
```

instrumentation/opentelemetry_grpc/README.md
```diff
@@ -0,0 +1,141 @@
...
+
+This library provides comprehensive tracing for gRPC applications, supporting both client and server instrumentation with proper context propagation according to OpenTelemetry semantic conventions.
+
+## Features
+
+- **Client Instrumentation**: Automatic tracing of outgoing gRPC requests
+- **Server Instrumentation**: Automatic tracing of incoming gRPC requests  
+- **Context Propagation**: Proper trace context propagation between services
+- **Error Handling**: Comprehensive error tracking and status reporting
```
sleipnir:
```suggestion
- **Error Handling**: Comprehensive error tracking and status reporting using gRPC status codes
```

instrumentation/opentelemetry_grpc/README.md
```diff
@@ -0,0 +1,141 @@
...
+
+```elixir
+OpentelemetryGrpc.Client.setup()
+```
+
+For context propagation in client requests, add the interceptor to your gRPC channel:
+
+```elixir
+{:ok, channel} = GRPC.Stub.connect("localhost:50051",
+  interceptors: [OpentelemetryGrpc.Interceptor])
```
sleipnir: I have a question here. If we have more interceptors, would this one need to be the first on the list? I mean, so that the context is propagated properly?

instrumentation/opentelemetry_grpc/README.md
```diff
@@ -0,0 +1,141 @@
...
+
+### Server Instrumentation
+
+To instrument only gRPC server requests:
+
+```elixir
+OpentelemetryGrpc.Server.setup()
+```
+
+By default, server instrumentation includes context propagation from incoming request headers. To control span relationships:
```
sleipnir, line 67: Here I think it's important to clarify when to use each type of relationship (:child, :link, etc.). For example:

* **:child**: Use when spans should be part of the same trace (default).
* **:link**: Use for fire-and-forget or detached async processing.
* **:none**: Disable context propagation (useful for isolated traces).

instrumentation/opentelemetry_grpc/README.md
```diff
@@ -0,0 +1,141 @@
...
+### Client-specific Attributes
+- `server.address`: Server hostname (when available)
+- `server.port`: Server port (when available)
+
+### Server-specific Attributes
+- `elixir.grpc.server_module`: The Elixir module handling the request
+- `elixir.grpc.endpoint_module`: The gRPC endpoint module
+
+### Error Attributes
+- `error.type`: Type of error that occurred
```
sleipnir:
```suggestion
- `error.type`: Type of error that occurred (e.g., GRPC.RPCError).
```

instrumentation/opentelemetry_grpc/README.md
```diff
@@ -0,0 +1,141 @@
...
+
+## Requirements
+
+- Elixir 1.11 or later
+- OpenTelemetry libraries for tracing setup
+- gRPC library for actual gRPC functionality
+
+## License
+
+Apache License 2.0
```
sleipnir:
```suggestion
Apache License 2.0 – see [LICENSE](LICENSE) for details.
```

instrumentation/opentelemetry_grpc/lib/opentelemetry_grpc/client.ex
```diff
@@ -0,0 +1,195 @@
+defmodule OpentelemetryGrpc.Client do
+  @moduledoc """
+  OpenTelemetry handler for gRPC client telemetry events.
+
+  This module handles client-side gRPC telemetry events and creates
+  OpenTelemetry spans for outgoing gRPC calls.
```
sleipnir:
```suggestion
  This module attaches telemetry handlers for gRPC client events, automatically creating OpenTelemetry spans for outgoing RPC calls following OpenTelemetry semantic conventions.
```

instrumentation/opentelemetry_grpc/lib/opentelemetry_grpc/client.ex
```diff
@@ -0,0 +1,195 @@
...
+  @doc """
+  Set up telemetry handlers for gRPC client events.
+
+  ## Examples
+
+      # Basic setup
+      OpentelemetryGrpc.Client.setup()
+
+  """
+  def setup do
```
sleipnir: it would be nice to add a spec (@spec) for these functions

instrumentation/opentelemetry_grpc/lib/opentelemetry_grpc/client.ex
```diff
@@ -0,0 +1,195 @@
...
+    OpentelemetryTelemetry.start_telemetry_span(@grpc_client_tracer_id, span_name, metadata, %{
+      kind: :client,
+      attributes: attributes
+    })
+  end
+
+  def handle_event(
+        [:grpc, :client, :rpc, :stop],
+        _measurements,
+        %{result: result} = metadata,
```
sleipnir: any chance of function clause error here?

instrumentation/opentelemetry_grpc/lib/opentelemetry_grpc/client.ex
```diff
@@ -0,0 +1,195 @@
...
+
+    otel_status = grpc_status_to_otel_status(status)
+    status_name = GRPC.Status.code_name(status)
+
+    Tracer.set_status(otel_status, message || "gRPC error: #{status_name}")
+
+    if status != 0 do
+      Tracer.add_event("rpc.message", %{
+        RPCAttributes.rpc_message_type() => RPCAttributes.rpc_message_type_values().received,
+        RPCAttributes.rpc_message_id() => 1,
```
sleipnir: hard coded?

sleipnir (changes requested):

instrumentation/opentelemetry_grpc/lib/opentelemetry_grpc/client.ex
yordis: removed for now, this would require headers or something like that, so we can add it eventually

instrumentation/opentelemetry_grpc/README.md
yordis: In practice, it shouldn't matter `as long as` people aren't dropping the header, or for some weird reasoning creating different context; when in doubt, at the end of the pipeline.

Now, I wonder, should I rename this to be called `ContextPropagatorInterceptor`? Since it could be confused by more than what it actually does.

I leave it to you and @tsloughter on that

instrumentation/opentelemetry_grpc/lib/opentelemetry_grpc/client.ex
yordis: the chance isn't 0, but hopefully it doesn't ... I am not sure to what extend we need to do defensive programming. I am following similar patterns like the rest.

instrumentation/opentelemetry_grpc/README.md
sleipnir: > In practice, it shouldn't matter `as long as` people aren't dropping the header, or for some weird reasoning creating different context; when in doubt, at the end of the pipeline.
>

Ok. So these concerns should be documented, that's my only point here.

> Now, I wonder, should I rename this to be called ContextPropagatorInterceptor? Since it could be confused by more than what it actually does.
>

`ContextPropagatorInterceptor` for me is a good name.

instrumentation/opentelemetry_grpc/lib/opentelemetry_grpc/client.ex
sleipnir: I find clause errors terribly annoying and prefer clear messages for users. I know this isn't always predictable everywhere, but I'd try to avoid it here if possible. But it's not a serious issue, so I'll leave it to your discretion.

sleipnir (approved):
LGTM

yordis:
@sleipnir do you mind giving it another code review; by now

instrumentation/opentelemetry_grpc/lib/opentelemetry_grpc/server.ex
```diff
@@ -0,0 +1,266 @@
...
+    Tracer.set_attributes(%{
+      RPCAttributes.rpc_grpc_status_code() => @status_code_internal,
+      ErrorAttributes.error_type() => to_string(error_struct)
+    })
+
+    Tracer.set_status(OpenTelemetry.status(:error, Exception.message(error)))
+  end
+
+  defp record_server_span_result(_result) do
+    # Unknown result format - don't set status
```
sleipnir, line 199: Just one question here, wouldn't it be interesting to return an error status?

instrumentation/opentelemetry_grpc/lib/opentelemetry_grpc/server.ex
```diff
@@ -0,0 +1,266 @@
...
+  defp setup_context_propagation(_source, _) do
+    []
+  end
+
+  defp extract_and_attach(source) do
+    case get_propagated_ctx(source) do
+      {_links, parent_ctx} when parent_ctx != :undefined ->
+        OpenTelemetry.Ctx.attach(parent_ctx)
+
+        # When we attach the context, we don't need links - parent-child relationship is established
```
sleipnir, line 228: It's clear to me why this is so, perhaps the comment is unnecessary

instrumentation/opentelemetry_grpc/lib/opentelemetry_grpc/server.ex
```diff
@@ -0,0 +1,266 @@
...
+        []
+
+      {links, _undefined_ctx} ->
+        # No parent context to attach, but we can still return links if any
```
sleipnir, line 232: Same as above

instrumentation/opentelemetry_grpc/lib/opentelemetry_grpc/server.ex
```diff
@@ -0,0 +1,266 @@
...
+    links
+  end
+
+  defp get_propagated_ctx(stream) do
+    stream
+    |> get_grpc_headers()
+    |> extract_to_ctx()
+  end
+
+  defp extract_to_ctx([]) do
```
sleipnir, line 248: Will an empty array or the presence of items always be returned? Any case clause risk?

instrumentation/opentelemetry_grpc/lib/opentelemetry_grpc/server.ex
```diff
@@ -0,0 +1,266 @@
...
+    {[], :undefined}
+  end
+
+  defp extract_to_ctx(headers) do
+    try do
+      ctx =
+        OpenTelemetry.Ctx.new()
+        |> :otel_propagator_text_map.extract_to(headers)
+
+      # Extract span context to check if it's valid and for creating links
```
sleipnir: unnecessary comment the intention is clear

instrumentation/opentelemetry_grpc/lib/opentelemetry_grpc/server.ex
```diff
@@ -0,0 +1,266 @@
...
+      span_ctx = OpenTelemetry.Tracer.current_span_ctx(ctx)
+
+      case span_ctx do
+        :undefined ->
+          # No valid parent span - no relationship possible
```
sleipnir: same as above

instrumentation/opentelemetry_grpc/lib/opentelemetry_grpc/server.ex
```diff
@@ -0,0 +1,266 @@
...
+          {[], :undefined}
+
+        span_ctx ->
+          # Return links first, then context (for parent-child relationships)
```
sleipnir: here I think the comment is valid

sleipnir (commented):
Just minor comments, we're getting there, congratulations.

What do you think about adding a small benchmark, comparing the interceptor for client and server? If it's not too much work, of course.

instrumentation/opentelemetry_grpc/lib/opentelemetry_grpc/server.ex
yordis: I think it should probably become a `Logger.warning` I am gonna wait to see what @tsloughter says
I am unsure of what we should do; in the internal one, I am using a logger here I think, but that means we must fix whatever happens to remove the warning

instrumentation/opentelemetry_grpc/lib/opentelemetry_grpc/server.ex
yordis: The funny thing is, I actually added this to help myself for future coding where I am gonna need to do the exact same thing, and then, it will no be clear, I will get confused on such triviality, until I relearn again 😄 

By now, I wish we have some helper functions around this topic, I am getting old, my brain is no braining like before 😭

instrumentation/opentelemetry_grpc/lib/opentelemetry_grpc/server.ex
```diff
@@ -0,0 +1,266 @@
...
+          {[OpenTelemetry.link(span_ctx)], ctx}
+      end
+    rescue
+      _error ->
+        {[], :undefined}
```
yordis: @tsloughter should I do this at all?

instrumentation/opentelemetry_grpc/lib/opentelemetry_grpc/server.ex
yordis: let me follow up on this tomorrow, maybe I shouldnt short-curcuit here, now thinking if we still want to do the linking stuff which now I am not doing 😓

yordis:
@sleipnir thank you so much for the review; I am gonna wait for @tsloughter input to tackle the comments and other things, some of them are there to actually help me to make sense of things, I honestly keep getting confused with Linking vs Child; and a nice pattern to follow

instrumentation/opentelemetry_grpc/lib/opentelemetry_grpc/server.ex
tsloughter: I'm not sure what is going on here? Can't this just be a fall through for the above `case`?

tsloughter:
Thanks for reviewing @sleipnir

instrumentation/opentelemetry_grpc/lib/opentelemetry_grpc/server.ex
yordis: Like, should I try-rescue at all?

instrumentation/opentelemetry_grpc/lib/opentelemetry_grpc/server.ex
yordis: I will keep it for now

instrumentation/opentelemetry_grpc/lib/opentelemetry_grpc/server.ex
yordis: I will keep it for now

instrumentation/opentelemetry_grpc/lib/opentelemetry_grpc/server.ex
yordis: @tsloughter do you have any take on this? As far as I can tell OTEL likes to leave things "unset", meaning don't do anything.

So it comes down to, should I log it or not?

instrumentation/opentelemetry_grpc/lib/opentelemetry_grpc/server.ex
yordis: The short-circuit on empty headers is correct and intentional. Here's why:

**What happens with empty headers:**
```elixir
defp extract_to_ctx([]) do
  {[], :undefined}
end
```

When `headers = []`:
1. There's no propagated context to extract (no traceparent, tracestate, etc.)
2. `:otel_propagator_text_map.extract_to/2` would have nothing to extract anyway
3. The result would be an empty context with `current_span_ctx(ctx) == :undefined`
4. We'd still return `{[], :undefined}`

instrumentation/opentelemetry_grpc/lib/opentelemetry_grpc/server.ex
yordis: other modules arent try-rescue so I am removing it

instrumentation/opentelemetry_grpc/lib/opentelemetry_grpc/server.ex
tsloughter: Why would it ever not match the above structures? Is it really typed to allow that or would it not mean a bug in the grpc library if it didn't match one of those?

yordis: It would be, most like, a bug in their end, so I decided no to crash the handler (since it woudlnt come back if that happens)

What do you want me to do here then?

sleipnir: Just to contribute, I don't believe there's a bug in the grpc library regarding this; we would probably have caught it in our tests by now. However, I think the approach of avoiding a handler failure here is sound, as a bug could be introduced there at some point anyway.

yordis: >  bug could be introduced there at some point anyway.

That is why I did it, in the end, it is about "level of trust" so if I remove it, somebody from the gRPC team may not know about the fact that "internal" data structures aren't that internal

tsloughter: Right, we are extra cautious in otel to avoid any chance of a crash, as well as not being accidentally noisy. So I guess we just leave this as is for now.

instrumentation/opentelemetry_grpc/lib/opentelemetry_grpc/client.ex
```diff
@@ -0,0 +1,166 @@
...
+    # The 3-tuple format returned when return_headers: true option is used.
+    record_client_span_result({:ok, response})
+  end
+
+  defp record_client_span_result({:ok, _response}) do
+    Tracer.set_attributes(%{
+      RPCAttributes.rpc_grpc_status_code() => @status_codes.ok
+    })
+
+    Tracer.set_status(OpenTelemetry.status(:ok))
```
tsloughter: I don't think this is right. Pretty sure you should leave this unset. Unless there is something in the spec saying otherwise for grpc?

instrumentation/opentelemetry_grpc/mix.exs
```diff
@@ -0,0 +1,88 @@
+defmodule OpentelemetryGrpc.MixProject do
+  use Mix.Project
+
+  @version "1.0.0"
```
tsloughter:
```suggestion
  @version "0.1.0"
```

Lets not get ahead of ourselves :). Let some others try it out and others compare against expected based on the spe cand then we can bump it.

sleipnir:
Thank you guys @tsloughter @yordis

instrumentation/opentelemetry_grpc/lib/opentelemetry_grpc/client.ex
```diff
@@ -0,0 +1,159 @@
...
+  @doc false
+  def handle_event(
+        [:grpc, :client, :rpc, :exception],
+        _measurements,
+        %{kind: kind, reason: reason, stacktrace: stacktrace} = metadata,
+        _config
+      ) do
+    span_ctx = OpentelemetryTelemetry.set_current_telemetry_span(@grpc_client_tracer_id, metadata)
+
+    OpenTelemetry.Span.record_exception(span_ctx, kind, reason, stacktrace)
```
grzuy, line 109: I don't think this function call would work?

We will need something like this in `opentelemetry_api` for it to work I think https://github.com/open-telemetry/opentelemetry-erlang/pull/965.

instrumentation/opentelemetry_grpc/lib/opentelemetry_grpc/server.ex
```diff
@@ -0,0 +1,284 @@
...
+    OpentelemetryTelemetry.set_current_telemetry_span(@grpc_server_tracer_id, metadata)
+    record_server_span_result(metadata.result)
+    OpentelemetryTelemetry.end_telemetry_span(@grpc_server_tracer_id, metadata)
+  end
+
+  @doc false
+  def handle_event([:grpc, :server, :rpc, :exception], _measurements, metadata, _config) do
+    ctx = OpentelemetryTelemetry.set_current_telemetry_span(@grpc_server_tracer_id, metadata)
+
+    OpenTelemetry.Span.record_exception(ctx, metadata.kind, metadata.reason, metadata.stacktrace)
```
grzuy, line 131: Same here.

PR #563, opened 2025-11-25
fix(opentelemetry_oban): update error handling in plugin to use formatted exception messages

yordis:
Signed-off-by: Yordis Prieto <yordis.prieto@gmail.com>

## Context

I saw the following log:

```
error: Handler "Elixir.OpentelemetryOban.PluginHandler.plugin_exception" has failed and has been detached. Class=:error
Reason=:function_clause
Stacktrace=[
  {Exception, :message, [:badarg], [file: ~c"lib/exception.ex", line: 92]},
  {OpentelemetryOban.PluginHandler, :handle_plugin_exception, 4,
   [file: ~c"lib/opentelemetry_oban/plugin_handler.ex", line: 64]},
  {:telemetry, :"-execute/3-fun-0-", 4,
   [file: ~c"/opt/umbrella/deps/telemetry/src/telemetry.erl", line: 167]},
  {:lists, :foreach_1, 2, [file: ~c"lists.erl", line: 2310]},
  {:telemetry, :span, 3,
   [file: ~c"/opt/umbrella/deps/telemetry/src/telemetry.erl", line: 346]},
  {Oban.Stager, :handle_info, 2, [file: ~c"lib/oban/stager.ex", line: 58]},
  {:gen_server, :try_handle_info, 3, [file: ~c"gen_server.erl", line: 2345]},
  {:gen_server, :handle_msg, 6, [file: ~c"gen_server.erl", line: 2433]}
]
````

### DIFF
--- a/instrumentation/opentelemetry_oban/lib/opentelemetry_oban/plugin_handler.ex
+++ b/instrumentation/opentelemetry_oban/lib/opentelemetry_oban/plugin_handler.ex
@@ -54,14 +54,21 @@ defmodule OpentelemetryOban.PluginHandler do
   def handle_plugin_exception(
         _event,
         _measurements,
-        %{kind: :error, reason: exception, stacktrace: stacktrace} = metadata,
+        %{kind: kind, reason: reason, stacktrace: stacktrace} = metadata,
         _config
       ) do
     ctx = OpentelemetryTelemetry.set_current_telemetry_span(@tracer_id, metadata)
 
     # Record exception and mark the span as errored
-    Span.record_exception(ctx, exception, stacktrace)
-    Span.set_status(ctx, OpenTelemetry.status(:error, Exception.message(exception)))
+    # We use :otel_span.record_exception/5 (Erlang) instead of Span.record_exception/3 (Elixir)
+    # because the Elixir version only works with exception structs (e.g., %RuntimeError{}),
+    # while the Erlang version handles ANY error reason (atoms like :badarg, tuples, etc.)
+    :otel_span.record_exception(ctx, kind, reason, stacktrace, [])
+
+    Span.set_status(
+      ctx,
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: instrumentation/opentelemetry_oban/lib/opentelemetry_oban/plugin_handler.ex [hunk 1]
```diff
@@ -54,14 +54,21 @@ defmodule OpentelemetryOban.PluginHandler do
...
       ) do
     ctx = OpentelemetryTelemetry.set_current_telemetry_span(@tracer_id, metadata)
 
     # Record exception and mark the span as errored
-    Span.record_exception(ctx, exception, stacktrace)
-    Span.set_status(ctx, OpenTelemetry.status(:error, Exception.message(exception)))
+    # We use :otel_span.record_exception/5 (Erlang) instead of Span.record_exception/3 (Elixir)
+    # because the Elixir version only works with exception structs (e.g., %RuntimeError{}),
+    # while the Erlang version handles ANY error reason (atoms like :badarg, tuples, etc.)
+    :otel_span.record_exception(ctx, kind, reason, stacktrace, [])
```
grzuy, line 66: For what is worth

I think there is a smaller change that also fixes the function clause error you are hitting, which is just adding

```elixir
exception = Exception.normalize(reason)
```
at the top of the function block and then keep the rest of the function unchanged.

The current diff seems to be doing more than fixing the issue. This is also handling `:throw` and `:exit` which they weren't being handled before.

Might be okay, but just flagging this to be aware of it.

yordis: I think the original code was suboptimal if you follow other packages in the repo, unless I made a mistake. Please create a PR about it.

PR #568, opened 2025-12-03
fix(opentelemetry_cowboy): prevent telemetry handler crash on invalid HTTP/2 scheme

### LINKED ISSUE #567: opentelemetry_cowboy crashes and detaches on malformed HTTP/2 :scheme pseudo-header
## Description

The `opentelemetry_cowboy` telemetry handler crashes when processing an HTTP/2 request with a malformed `:scheme` pseudo-header. Once the handler crashes, the Erlang `telemetry` library permanently detaches it, causing all subsequent requests on that node to lose OpenTelemetry tracing.

## Root Cause

In [`opentelemetry_cowboy.erl:239`](https://github.com/open-telemetry/opentelemetry-erlang-contrib/blob/07ec3d60093be802d63d20d7a82e87e7a7993950/instrumentation/opentelemetry_cowboy/src/opentelemetry_cowboy.erl#L239-L244), the `extract_scheme/2` function only handles `<<"http">>` and `<<"https">>`:

```erlang
case ReqScheme of
    <<"http">> ->
        http;
    <<"https">> ->
        https
end;
```

Any other value causes a `case_clause` crash. Cowboy does not validate the `:scheme` pseudo-header and passes through arbitrary values, so malformed requests (from scanners, bots, or misconfigured proxies) can trigger this crash.

## Reproduction

Minimal reproduction repository: https://github.com/velimir/scheme_crash

```bash
git clone https://github.com/velimir/scheme_crash
cd scheme_crash
mix deps.get
iex -S mix
```

In another terminal:
```bash
nghttp -v -n -H':scheme: http://example.com/?' http://localhost:8080/
```

## Error Output

```
[error] Handler {:opentelemetry_cowboy, :otel_cowboy} has failed and has been detached. Class=:error
Reason={:case_clause, "http://example.com/?"}
Stacktrace=[
  {:opentelemetry_cowboy, :extract_scheme, 2,
   [file: '.../opentelemetry_cowboy.erl', line: 239]},
  {:opentelemetry_cowboy, :handle_event, 4,
   [file: '.../opentelemetry_cowboy.erl', line: 295]},
  ...
]
```

## Versions

- `opentelemetry_cowboy`: 1.0.0
- `cowboy`: 2.14.2
- `opentelemetry`: 1.7.0
- Erlang/OTP: 27.x

velimir:
Cowboy does not validate the `:scheme` pseudo-header per RFC 9113, which
states that scheme "is not restricted to http and https" and can be used
for non-HTTP schemes via proxies/gateways. This means Cowboy passes through
any value from the HTTP/2 HEADERS frame without validation.

When a malformed or unexpected scheme value reaches `extract_scheme/2`,
the case clause crashes because it only handles `<<"http">>` and
`<<"https">>`. This crash causes Erlang's telemetry library to permanently
detach the handler for the affected node, resulting in all subsequent
requests losing OpenTelemetry tracing until the node is restarted.

In production, this manifests as "almost all" requests suddenly missing
trace IDs, with the root cause being a single malformed HTTP/2 request
that crashed the telemetry handler.

The fix introduces a configurable `valid_schemes` option that maps scheme
binaries to atoms. Unknown schemes now return `undefined` instead of
crashing, which is consistent with how `otel_http:extract_scheme/2` already
handles unknown schemes from headers.

The default configuration preserves backwards compatibility by only
accepting `http` and `https`. Users who need to support additional schemes
(e.g., `ws`, `wss`) can extend the map via configuration.

Fixes: #567


instrumentation/opentelemetry_cowboy/test/opentelemetry_cowboy_SUITE.erl
```diff
@@ -507,3 +508,20 @@ binary_status_code_request(_Config) ->
...
+    {ok, {{_Version, 200, _ReasonPhrase}, _Headers, _Body}} =
+        httpc:request(get, {"http://localhost:8080/success", []}, [], [{socket_opts, [{port, Port}]}]),
+    receive
+        {span, #span{name=Name,attributes=Attributes,kind=Kind}} ->
+            ?assertEqual('GET', Name),
+            ?assertEqual(?SPAN_KIND_SERVER, Kind),
+            ?assertEqual(undefined, maps:get(?URL_SCHEME, otel_attributes:map(Attributes)))
+    after
+        1000 -> ct:fail(invalid_scheme_returns_undefined)
+    end.
```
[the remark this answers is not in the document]
velimir: I would love to, but making an http2 request via `httpc` is.. problematic.

velimir:
hey @tsloughter is there anything I can help you with to get this merged?

instrumentation/opentelemetry_cowboy/src/opentelemetry_cowboy.erl
```diff
@@ -95,7 +95,8 @@ default_opts() ->
         scheme_headers => [<<"forwarded">>, <<"x-forwarded-proto">>],
         scheme_headers_sort_fn => undefined,
         server_address_headers => [<<"forwarded">>, <<"x-forwarded-host">>, <<"host">>],
-        server_address_headers_sort_fn => undefined
+        server_address_headers_sort_fn => undefined,
+        valid_schemes => #{<<"http">> => http, <<"https">> => https}
```
tsloughter, line 99: If this is only matched on in 1 place can't we just put the logic there and not use a map but just match the http/https atoms in a case?

velimir: The idea is to allow users to configure the valid schemes that are relevant to their project; therefore, the map is located in the configuration.

Sorry, does this answer your question? Or are you wondering whether it would be better to just match and return undefined if those do not match either http or https in the `extract_scheme` function? If that is the case, then it would not be possible to have schemes like websocket supported, for example.

tsloughter: Websockets! I always forget websockets :). I knew there was something I was forgetting. But why aren't `ws` and `wss` included here?

But yea, I didn't catch that the user could supply their own. That makes sense.

velimir: idk, tbh, it's been like that and I didn't want to make any changes in the behaviour unless a user opts in 🤷

If you are happy with the approach I can also add the ws and wss into the default list. What do you think?

tsloughter: Yea, I guess just add the wss and ws. But wait, what did you see it crashing on?

tsloughter: Oh, invalid schemas? How were those getting through... interesting

velimir: > Cowboy does not validate the :scheme pseudo-header and passes through arbitrary values, so malformed requests (from scanners, bots, or misconfigured proxies) can trigger this crash.

There's a repo in the issue linked if you want to play with it ([issue](https://github.com/open-telemetry/opentelemetry-erlang-contrib/issues/567), [test project](https://github.com/velimir/scheme_crash))

instrumentation/opentelemetry_cowboy/src/opentelemetry_cowboy.erl
```diff
@@ -236,12 +239,7 @@ extract_scheme(Req, Config) ->
     SchemeHeaders1 = extract_headers(Headers, SchemeHeaders),
     case otel_http_extract_scheme(SchemeHeaders1, SortFn) of
         undefined ->
-            case ReqScheme of
-                <<"http">> ->
-                    http;
-                <<"https">> ->
-                    https
-            end;
+            maps:get(ReqScheme, ValidSchemes, undefined);
```
tsloughter, line 242: I just realized. Why does this even have to be an atom? Can't we just use `ReqScheme` and forget all this other stuff of validation and conversion?

velimir: Fair enough, let me work on this 👌

velimir: Done in 86e592c — removed `valid_schemes` entirely and pass `ReqScheme` binary through directly.

velimir:
@tsloughter this is ready for another look when you get a chance

tsloughter:
Hey, sorry, I should have reviewed closer the first time. We don't just want to return `ReqScheme` since that could be used to blow up the cardinality of the scheme attribute. The right fix, I think, is to update `otel_http` to support `ws` and `wss`.

yordis (approved):

PR #580, opened 2025-12-15
refactor(opentelemetry_broadway): remove JobHandler module to streamline telemetry integration

yordis:
This sounds like a copy+paste duplication


tsloughter:
Ah, so this isn't used anywhere?

yordis:
@tsloughter maybe @whatyouhide could share some thoughts here, but I have never used that module, and as far as I can tell, it is duplicate of what is done in the main module.

whatyouhide (approved):

PR #602, opened 2026-02-23
docs: add configuration options reference

yordis:
## Summary

- Adds a reference document defining the standard configuration options (`opt_in_attrs`, `opt_out_attrs`, `extra_attrs`, `span_relationship`) for instrumentation packages
- Includes a mapping from OTel SemConv requirement levels to the corresponding options
- Includes a package support matrix showing which packages implement which options

## Test plan

- [x] Review document accuracy against existing package implementations
- [x] Verify SemConv requirement level mapping is correct


yordis:
@tsloughter can we take this one?

PR #621, opened 2026-03-06
Add http.route attribute and update span name from Plug router

lunks:
## Summary
- At request stop, reads `conn.private[:plug_route]` to build `"GET /items/:id"` span names and set the `http.route` attribute
- Non-standard HTTP methods are normalized to `"HTTP"` to prevent unbounded span name cardinality
- Works with Phoenix and any Plug-based router that sets `:plug_route`

Follows up on #401.


instrumentation/opentelemetry_bandit/lib/opentelemetry_bandit.ex
```diff
@@ -468,10 +468,22 @@ defmodule OpentelemetryBandit do
       |> Tracer.set_attributes()
     end
 
+    maybe_update_span_name(conn)
```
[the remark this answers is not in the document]
lunks: If reviewers think we should address it, let me know. Resolving it isn't worth it, and verifying at the end is intentional.

instrumentation/opentelemetry_bandit/lib/opentelemetry_bandit.ex
```diff
@@ -468,10 +468,29 @@ defmodule OpentelemetryBandit do
...
+
     Tracer.end_span()
     Ctx.clear()
   end
 
+  defp maybe_update_span_name(conn) do
+    case conn.private do
+      %{plug_route: {route, _fun}} ->
+        method = sanitize_method(conn.method)
+        Tracer.update_name("#{method} #{route}")
```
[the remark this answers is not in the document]
lunks: Marking as resolved, I don't believe this is necessary.

instrumentation/opentelemetry_bandit/lib/opentelemetry_bandit.ex
```diff
@@ -264,7 +264,7 @@ defmodule OpentelemetryBandit do
 
   # error with no conn
   def handle_request_start(_meta, _config) do
-    Tracer.start_span(:HTTP, %{kind: :server})
+    Tracer.start_span("HTTP", %{kind: :server})
```
yordis: how does those this affect the final span name?

instrumentation/opentelemetry_bandit/lib/opentelemetry_bandit.ex
```diff
@@ -468,10 +468,30 @@ defmodule OpentelemetryBandit do
       |> Tracer.set_attributes()
     end
 
+    maybe_update_span_name(conn)
```
yordis, line 471: This couldn't be moved to the start event?

lunks: No. We only have the URL path, not the route name. `call_plug!` sets it, but the span has already been started:
https://github.com/mtrudel/bandit/blob/1.10.3/lib/bandit/pipeline.ex#L38-L42

yordis: @tsloughter I am gonna defer to you, I am not sure if changing midway the span name is a major concern or not

lunks: If you folks think there's a better approach to accomplish a similar result, let me know.

tsloughter-serve: Sorry. This is fine.

yordis (approved):

PR #467, opened 2025-02-17
fix(bandit): handles bandit throw/exit events

grzuy:
~Testing that it doesn't receive a span because currently testing against bandit 1.5 (https://github.com/mtrudel/bandit/pull/443 added in bandit 1.6.2).~

~Still pending testing against bandit 1.6 (https://github.com/open-telemetry/opentelemetry-erlang-contrib/pull/458)~

- Updates tested bandit to latest (1.6.8)
- Fixes handling of throw/exit events (added in bandit 1.6.2)

supersedes https://github.com/open-telemetry/opentelemetry-erlang-contrib/pull/458

PR #646, opened 2026-03-18
feat(bandit): record exception events for throws and abnormal exits too

grzuy:
Hi,

Following up on https://github.com/open-telemetry/opentelemetry-erlang-contrib/pull/467.

That one originally fixed `opentelemetry_bandit` to not crash when a throw/exit was emitted by `bandit`, and properly register the span with the correct error status.

BUT we never actually also recorded the OTel exception event for the case `bandit` emits a throw or a exit telemetry exception event.

This PR tries to improve that.

Thanks.


tsloughter (approved):

yordis:
https://github.com/open-telemetry/opentelemetry-erlang-contrib/pull/658 @tsloughter @grzuy can you please review the PR, ideally we cover the 95% of the cases to the point we are copy+pasting

PR #647, opened 2026-03-18
test(oban): simplifies running tests by not depending on postgres

grzuy:
Refactors `opentelemetry_oban` setup so that it doesn't depend on having a PostgreSQL instance running locally.

Using sqlite3 like this should "just work" for anyone trying to run `mix test` locally.


yordis:
Given that the Oban code barely changes, so running on CI or locally isn't a big deal, personally speaking.

I wonder,

Is it a good idea to optimize it when, in practice, Postgres is most likely the bulk of usage, and some accidental mismatching of behavior would make the situation worse?

grzuy:
Gotcha, thanks for the feedback.

PR #696, opened 2026-05-14
chore(renovate): batch and slow down dependency updates

yordis:
- Renovate currently opens far too many PRs, one per dependency per subproject, which drowns out signal and burns CI.
- Move to a monthly cadence so reviewers get a predictable, batched window instead of constant churn — CVEs still bypass the schedule.
- Auto-group updates per subproject via `{{parentDir}}` so new instrumentation/propagator/utility directories are covered without editing this config again.
- Cap concurrent PRs to keep the queue reviewable.

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -4,11 +4,35 @@
     "config:best-practices",
     "helpers:pinGitHubActionDigestsToSemver"
   ],
+  "labels": ["renovate"],
+  "schedule": ["before 6am on monday"],
+  "prConcurrentLimit": 10,
+  "prHourlyLimit": 0,
+  "vulnerabilityAlerts": {
+    "schedule": ["at any time"]
+  },
   "packageRules": [
     {
+      "matchUpdateTypes": ["patch", "pin", "digest"],
+      "groupName": "patch updates",
+      "addLabels": ["patch", "chore", "skip-changelog"]
+    },
+    {
       "matchDatasources": ["hex"],
-      "matchUpdateTypes": ["minor", "patch", "pin", "digest"],
-      "addLabels": ["skip-changelog", "chore"]
+      "matchUpdateTypes": ["minor"],
+      "groupName": "hex minor updates",
+      "addLabels": ["minor", "chore"]
+    },
+    {
+      "matchDatasources": ["hex"],
+      "matchUpdateTypes": ["major"],
+      "dependencyDashboardApproval": true,
+      "addLabels": ["major"]
+    },
+    {
+      "matchManagers": ["github-actions"],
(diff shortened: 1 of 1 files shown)


renovate.json
```diff
@@ -1,13 +1,34 @@
...
+    },
+    {
+      "matchFileNames": [
+        "instrumentation/*/**",
+        "propagators/*/**",
+        "utilities/*/**"
+      ],
+      "groupName": "{{parentDir}} deps",
+      "additionalBranchPrefix": "{{parentDir}}-"
+    },
```
thompson-tomo, line 31: This will create a significant bottleneck. For instance if a single hex update is available for 15 libraries that would take 5 months to cycle through assuming no other updates but in reality you will get atleast a ci update a month meaning it can now take 8 months or even more.

renovate.json
```diff
@@ -1,13 +1,34 @@
 {
   "$schema": "https://docs.renovatebot.com/renovate-schema.json",
-  "extends": [
-    "config:recommended"
-  ],
+  "extends": ["config:recommended"],
+  "schedule": ["* * 1-7 * 1"],
+  "prConcurrentLimit": 3,
```
thompson-tomo: This combination will be too low & infrequent to allow you to keep dependencies updated.

thompson-tomo (commented):
What i would propose instead based on what a number of sigs do especially as the proposal here will create a bottleneck is:

- have a single group for patch updates
- have another group for minor hex updates
- require major hex to have an approval before being raised

Introduce a labeller to indicate on the pr what packages are being updated.

I would keep the frequency at weekly as if it takes 2 weeks to process, the pr will update as needed.

This way you end up with 2 hex updates & maybe a couple ci updates.

yordis:
@thompson-tomo I changed few things

renovate.json
```diff
@@ -1,13 +1,46 @@
...
+      "matchFileNames": ["instrumentation/*/**"],
+      "addLabels": ["instrumentation"]
+    },
+    {
+      "matchFileNames": ["propagators/*/**"],
+      "addLabels": ["propagators"]
+    },
+    {
+      "matchFileNames": ["utilities/*/**"],
+      "addLabels": ["utilities"]
```
thompson-tomo:
```suggestion
```
Should already be covered by https://github.com/open-telemetry/opentelemetry-erlang-contrib/blob/main/.github/labeler.yml hence un-necessary.

thompson-tomo:
@yordis i am much happier with this latest iteration and it more closely aligns with the objective.

bryannaegele (approved):

Issue #322, opened 2024-05-01
DBConnection.ConnectionError causing Oban telemetry handler to detach


davydog187:
**Describe the bug**
Seems like there is a case where the Oban telemetry handler doesn't handle a reraised exception from DBConnection

```elixir
[error] Handler "Elixir.OpentelemetryOban.PluginHandler.plugin_exception" has failed and has been detached. Class=:error
Reason=:function_clause
Stacktrace=[
  {OpentelemetryOban.PluginHandler, :handle_plugin_exception,
   [
     [:oban, :plugin, :exception],
     %{monotonic_time: -576452278018730500, duration: 7500619507},
     %{
       reason: %DBConnection.ConnectionError{
         message: "connection not available and request was dropped from queue after 608ms. This means requests are coming in and your connection pool cannot serve them fast enough. You can address this by:\n\n  1. Ensuring your database is available and that you can connect to it\n  2. Tracking down slow queries and making sure they are running fast enough\n  3. Increasing the pool_size (although this increases resource consumption)\n  4. Allowing requests to wait longer by increasing :queue_target and :queue_interval\n\nSee DBConnection.start_link/2 for more information\n",
         severity: :error,
         reason: :queue_timeout
       },
       stacktrace: [
         {DBConnection, :transaction, 3,
          [
            file: ~c"lib/db_connection.ex",
            line: 1059,
            error_info: %{module: Exception}
          ]},
         {Oban.Plugins.Pruner, :"-handle_info/2-fun-0-", 2,
          [file: ~c"lib/oban/plugins/pruner.ex", line: 109]},
         {:telemetry, :span, 3,
          [
            file: ~c"/truncateddeps/telemetry/src/telemetry.erl",
            line: 321
          ]},
         {Oban.Plugins.Pruner, :handle_info, 2,
          [file: ~c"lib/oban/plugins/pruner.ex", line: 108]},
         {:gen_server, :try_handle_info, 3,
          [file: ~c"gen_server.erl", line: 1095]},
         {:gen_server, :handle_msg, 6, [file: ~c"gen_server.erl", line: 1183]},
         {:proc_lib, :init_p_do_apply, 3, [file: ~c"proc_lib.erl", line: 241]}
       ],
       conf: %Oban.Config{
         dispatch_cooldown: 5,
         engine: Oban.Pro.Engines.Smart,
         get_dynamic_repo: nil,
         insert_trigger: true,
         log: false,
         name: Oban,
[...]

PR #698, opened 2026-06-25
fix(opentelemetry_oban): guard nil :jobs in Cron plugin stop attributes

patrols:
## Problem

`OpentelemetryOban.PluginHandler` crashes with `:badarg` when an
`Oban.Plugins.Cron` (or `Oban.Pro.Plugins.DynamicCron`) `[:oban, :plugin, :stop]`
event carries no `:jobs` in its metadata:

```
Handler "Elixir.OpentelemetryOban.PluginHandler.plugin_stop" has failed and has been detached.
Class=:error Reason=:badarg
  {:erlang, :length, [nil], ...}
  {OpentelemetryOban.PluginHandler, :end_span_plugin_attrs, 1, [line: 87]}
```

Because `:telemetry` detaches any handler that raises, this **silently disables
OpentelemetryOban plugin tracing on that node until the next restart**.

## Root cause

`Oban.Plugins.Cron` only adds `:jobs` to the stop metadata on the success path.
When the scheduled insert fails — e.g. a transient DB error — it returns
`{:error, Map.put(meta, :error, error)}` with **no** `:jobs` key. `end_span_plugin_attrs/1`
then evaluates `length(metadata[:jobs])` = `length(nil)` → `:badarg`.

This is the same DB-error scenario reported in #322. #392 fixed the related
`:function_clause` crash on that path but left this `length(nil)` `:badarg` in the
`Cron`/`DynamicCron` attribute builders.

## Fix

Guard the two `length/1` calls with `metadata[:jobs] || []`, so `jobs_count`
defaults to `0` instead of crashing.

## Tests

Adds regression tests for the no-`:jobs` stop event for both `Oban.Plugins.Cron`
and `Oban.Pro.Plugins.DynamicCron`. Without the fix they reproduce the production
`:badarg` detach; with it the full suite passes (15/15 plugin handler, 42/42 overall).

---

Note: I verified locally, but `main` does not currently compile on Elixir 1.20.x due
to an unrelated `insert/5` & `insert_all/5` default-argument conflict
(`def insert/5 defaults conflicts with insert/3`) — 1.20 promotes this from a warning
to an error. Happy to open a separate issue/PR for that.

### DIFF
--- a/instrumentation/opentelemetry_oban/CHANGELOG.md
+++ b/instrumentation/opentelemetry_oban/CHANGELOG.md
@@ -1,5 +1,16 @@
 # Changelog
 
+## Unreleased
+
+### Fixed
+
+* Fix `OpentelemetryOban.PluginHandler` crashing with `:badarg` when an
+  `Oban.Plugins.Cron` (or `Oban.Pro.Plugins.DynamicCron`) `[:oban, :plugin, :stop]`
+  event carries no `:jobs` in its metadata. Oban omits `:jobs` when the scheduled
+  insert fails (e.g. a transient DB error), so `length(metadata[:jobs])` raised on
+  `nil` and `:telemetry` detached the handler, silently disabling plugin tracing
+  until the next restart. `jobs_count` now defaults to `0`.
+
 ## 1.1.1
 
 ### Fixed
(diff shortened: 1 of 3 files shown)


yordis:
hey @patrols do you mind signing the EasyCLA?

patrols:
@yordis sure, I have signed it now

yordis (approved):

PR #788, opened 2026-08-18
fix(opentelemetry_oban): report plugin failures and populate plugin counts

yordis:
Supersedes #195. That PR's implementation did not survive review, so this is a fresh branch off `main` rather than a rebase, but the original report and problem framing came from @andrewhr and the credit for finding this belongs to them.

- Oban plugins signal failure through the `{:error, meta}` return value of `:telemetry.span/3`, which emits a `:stop` event and not an `:exception` event. Every failing plugin run therefore arrived at the backend with an unset status, indistinguishable from a successful one, so plugin failures were invisible in exactly the situation where a trace is most useful.
- The one exception to the above is deliberate: a `nil` error is how `Oban.Plugins.Reindexer` reports the non-leader case, which happens on every follower node on every scheduled run. Treating that as a failure would have produced continuous false alarms.
- Plugin exception spans recorded no `exception.message`, which is the field a human actually reads first when triaging.
- The Lifeline and DynamicLifeline count attributes read metadata keys Oban does not emit, so they were never populated. Because the exporter serializes them anyway, they reached the backend as the literal string `"nil"`, which is worse than absent: it poisons any aggregation over those attributes.
- Job exception spans matched on a metadata key that Oban's documented job telemetry does not include, so the intended data was not being read at all.
- The plugin identity was emitted with the internal `Elixir.` prefix, which does not match the module name users write in their own Oban config, making spans hard to correlate with the code that produced them. This is a breaking change for anyone filtering on the attribute or grouping on the span name, and it is called out in the CHANGELOG.

### DIFF
--- a/instrumentation/opentelemetry_oban/CHANGELOG.md
+++ b/instrumentation/opentelemetry_oban/CHANGELOG.md
@@ -11,6 +11,48 @@
   `nil` and `:telemetry` detached the handler, silently disabling plugin tracing
   until the next restart. `jobs_count` now defaults to `0`.
 
+* Report the real discarded and rescued counts for `Oban.Plugins.Lifeline` and
+  `Oban.Pro.Plugins.DynamicLifeline`. Both read `:discarded_count` and
+  `:rescued_count`, which Oban does not emit; the plugins report the
+  `:discarded_jobs` and `:rescued_jobs` lists instead. The attributes were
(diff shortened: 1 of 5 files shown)


yordis:
@open-telemetry/erlang-contrib-approvers team, I need some 1+ here

tsloughter (approved):

PR #465, opened 2025-02-05
add http.route attribute. issue #464

sc-yan:
see detail at https://github.com/open-telemetry/opentelemetry-erlang-contrib/issues/464

PR #809, opened 2026-09-14
feat(opentelemetry_phoenix): add http.route to LiveView mount and handle_params spans

yordis:
- LiveView spans identify work only by the view module, but the router is the real entry point for a live route. One live module can serve several routes, and `inspect(socket.view)` collapses them into a single identity, so route-level latency and error rates are not separable today.
- `http.route` is what this instrumentation already emits from router dispatch (#465), so LiveView spans now agree with HTTP server spans instead of exposing a second, module-shaped convention for the same concept.
- Non-breaking on purpose. Span names are untouched and no existing attribute changes, which keeps the concerns raised on #433 and #432 out of scope.
- Scoped to `mount` and `handle_params` because those are the only LiveView telemetry events that carry a URI. `handle_event`, `render` and `live_component` `update` receive `%{socket: ..., ...}` with no URI, and `socket.host_uri` is pruned to scheme/host/port, so there is no honest route to report there. Extending coverage needs a route cache or an upstream LiveView change, which belongs in its own change.
- Guarded against the two cases where a route cannot be resolved: `Phoenix.LiveView.Static` mounts LiveViews that are not at the router with `uri: nil`, and such sockets can carry `router: nil`. Unguarded, both raise inside the telemetry handler, and telemetry then permanently detaches it. Since every LiveView event shares the handler id `{OpentelemetryPhoenix, :live_view}`, that would silently disable all LiveView tracing for the node. Tests cover both, plus an unmatched path and a parameterized route.
- `phoenix_live_view` is added as a `:dev`/`:test` dependency only, so the published package gains no dependency on the instrumented library.

- Verified against a real endpoint, not only recorded metadata. `test/live_view_integration_test.exs` boots a `Phoenix.Endpoint` with a real router and drives an actual request through it, so LiveView itself emits the telemetry. It asserts the route on `mount` and `handle_params`, and asserts that a nested `live_render` child, which LiveView mounts with no URI, is still traced and leaves the handler attached. Deleting the guards in `route_attributes/1` makes that suite fail with `Handler {OpentelemetryPhoenix, :live_view} has failed and has been detached`, while the existing fixture-based suite stays green.


instrumentation/opentelemetry_phoenix/test/support/router.ex
GregMefford, line 1: Is this file getting used for something, or was it temporary during development?

yordis: the module name has been there for years 😭

tsloughter (approved):

PR #810, opened 2026-09-15
feat(opentelemetry_phoenix): name LiveView spans after the route behind an opt-in

yordis:
- Naming every LiveView span after the module makes mounts of the same LiveView indistinguishable by route, which is the information most people want when they look at a LiveView trace.
- Renaming outright breaks existing dashboards, alerts and saved queries, and that is what stalled #433, so the route naming is opt-in through `liveview_span_names: :route` and the module naming stays the default.
- `handle_event` needed more than a rename to be useful: its metadata carries no `:uri` and the socket keeps no path, so the route resolved during `mount`/`handle_params` is carried forward for the lifetime of the LiveView process.
- Supersedes #431 and #433, which can be closed.


tsloughter:
Which is closer to what the spec requires? The default should reflect that.

yordis:
> Which is closer to what the spec requires? The default should reflect that.

@tsloughter wouldnt that require a breaking change thou? we should avoid such thing and maybe use "deprecate" labels?

yordis:
Semantic Conventions don't cover LiveView spans. [`docs/general/trace.md`](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/general/trace.md) enumerates the defined span conventions (General attributes, Compatibility, CloudEvents, Cloud Providers, Database, Exceptions, FaaS, HTTP, Messaging, Object Stores, RPC), and there is no framework or request-handler convention among them.

The only place `http.route` enters a span name is the HTTP convention, and it is scoped away from these spans. From [`http-spans.md#name`](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/http/http-spans.md#name):

> HTTP span names SHOULD be `{method} {target}` [...] The `{target}` SHOULD be one of the following: [`http.route`](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/registry/attributes/http.md) for HTTP Server spans
>
> The `{method}` MUST be `{http.request.method}` if the method represents the original method known to the instrumentation. In other cases (when `{http.request.method}` is set to `_OTHER`), `{method}` MUST be `HTTP`.

`mount`, `handle_params` and `handle_event` set no `http.request.method`, so there is no `{method}` half to build a name from. That rule governs the router span, which this library already names `"#{meta.conn.method} #{meta.route}"`, not the LiveView lifecycle spans.

The closest analogue Semantic Conventions actually define is GraphQL server spans: a framework-level operation, `SERVER` kind, no route. [`graphql-spans.md`](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/graphql/graphql-spans.md) lands on the conservative default with the descriptive name behind an option:

> **Span name** SHOULD be of the format `{graphql.operation.type}` provided `graphql.operation.type` is available.
>
> Instrumentation MAY provide a configuration option to enable a more descriptive span name following `{graphql.operation.type} {graphql.operation.name}` format when `graphql.operation.name` is available.

That is structurally what this PR does. In fairness, GraphQL's stated reason is the high cardinality of a client-supplied operation name, and that rationale does not transfer here, since route templates are low cardinality and bounded by the router.

Where the conventions do bear on this: wherever a span naming format exists, it is `{operation} {target}`, as in `{db.operation.name} {target}` and `{method} {target}`. That is why the opt-in produces `live_view.mount /resources/:resource_id` rather than a module name.

So I don't think the conventions settle the default. The reason `:module` is the default is compatibility, not conformance. If you'd rather `:route` be the default and treat the module names as the legacy mode, I'm happy to flip it, 2.0.0 already set the precedent for a breaking span name change.

tsloughter:
Ok. If the conventions don't settle the default then we can leave it as this PR has it.

For future reference, I believe we don't have to do a major bump if the change is to follow the spec or the conventions. But I'm also fine with lots of major bumps :)

tsloughter (approved):

PR #811, opened 2026-09-15
test(opentelemetry_phoenix): drive LiveView telemetry through real Phoenix

yordis:
- The LiveView tests asserted against hand-maintained telemetry metadata, which cannot show that the instrumentation works against the events Phoenix actually emits.
- That metadata had already drifted from LiveView: real metadata always carries `telemetry_span_context`, which selects a different span correlation path in `opentelemetry_telemetry` than the fixture exercised, so the mechanism under test was not the one that runs in production.
- Every event the instrumentation subscribes to is reachable from a real LiveView, including the exception paths and the `live_component` ones, so there is little left for the fixture to justify.
- What cannot be reached from a real LiveView is now the only thing left synthesized: a socket with no router, and a URI matching no route.
- No behavior change. The new assertions document what `main` does today, which should make the behavior change in a follow-up visible in its own diff.


yordis:
Prerequisite for #810, which adds the behavior these tests will then assert on.

GregMefford (approved):
Excellent. I was very confused on the previous PR why all that stuff was hard-coded and this makes a lot more sense to me as a way forward.

tsloughter (approved):

